PR #496, opened 2023-11-22
Implement functional substitutions in rule applications

MirceaS:


generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -319,34 +319,52 @@ def current_configuration(self) -> Pattern:
         """Returns the current configuration."""
         return self._curr_config
 
+    @staticmethod
```
IljaZakharov, line 322: We can keep it as is, or we can make them just independent functions because we will need them also in the `SimplificationProver` class.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -319,34 +319,52 @@ def current_configuration(self) -> Pattern:
...
+        for name, plug in substitution.items():
+            claim = claim.apply_esubst(name, plug)
+        self.add_claim(claim)
+        sort, lhs, rhs = kl.kore_rewrites.assert_matches(claim)
+
+        # Check that the lhs equals the current configuration
         assert (
             lhs == self.current_configuration
-        ), f'The current configuration {lhs.pretty(self.pretty_options())} does not match the lhs of the rule {rule.pattern.pretty(self.pretty_options())}'
+        ), f'The current configuration {str(self.current_configuration)} does not equal {str(lhs)}, the lhs of an instance of rule {str(rule.pattern)}'
```
IljaZakharov, line 343: I thought the `pretty` is preferable now, especially when the code is inside the proof expression.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -319,34 +319,52 @@ def current_configuration(self) -> Pattern:
...
 
-        # Add the axioms
-        self.add_assumptions_for_rewrite_step(rule, substitution)
-
-        # Add the claim
-        self.add_claim(instantiated_axiom)
+        # Add the axioms/assumptions
```
IljaZakharov, line 345: This block for adding functional axioms should be possible to use in both `SimplificationProver` and `SimplificationPerformer`. But we can extract it later.

generation/src/tests/unit/test_execution_proof_generation.py
```diff
@@ -81,6 +81,18 @@ def rewrite_hints() -> list[RewriteStepExpression]:
     return [step_one, step_two]
 
 
+# TODO: Add side conditions!
```
IljaZakharov, line 84: The function was moved to `test_kore_language_semantics`. It is called `rewrite_with_cells_config_pattern`.

generation/src/tests/unit/test_kore_language_semantics.py
```diff
@@ -600,38 +599,39 @@ def test_collect_functional_axioms() -> None:
         with module as mod:
             sort = mod.sort('sort')
             a_symbol = mod.symbol('a', sort, input_sorts=(sort,), is_functional=True, is_ctor=True)
```
IljaZakharov, line 601: It would be nice to have a separate function for checking assertions and doing functional substitutions, including adding functional axioms. We need the function in two classes. So it would be nice to have tests arranged correspondingly. But this can also be done in a separate PR as we will need to make other changes.

generation/src/tests/unit/test_kore_language_semantics.py
```diff
@@ -600,38 +599,39 @@ def test_collect_functional_axioms() -> None:
...
             c = c_symbol.app
+            d = d_symbol.app
+            single_evar_krule = mod.rewrite_rule(kore_rewrites(sort.aml_symbol, EVar(0), EVar(0)))
+            double_evar_krule = mod.rewrite_rule(
+                kore_rewrites(sort.aml_symbol, d(EVar(0), EVar(1)), d(EVar(0), EVar(1)))
+            )
 
-            axioms = ExecutionProofExp.collect_functional_axioms(
-                sem,
+            b_pe = ExecutionProofExp(sem, b)
```
IljaZakharov, line 614: Let's move it to the [test_kore_language_semantics.py](https://github.com/runtimeverification/proof-checker/pull/496/files#diff-e0683fa613aac0b14ad594d14cc30eb5fd3f1a6e367cde89cd4b6fd986ce46f3) as we use the `ExecutionProofExp` class. Another option is just to split the test.

IljaZakharov (approved):
Looks nice to me! I added a couple of comments, but they are minor, and we can address them in a separate PR because we will need to reuse the substitution logic in the `SimplificationInfo` class.

fiedlr (changes requested):
Let's not merge this until Nishant or I go over this, I want to doublecheck the proof sizes.

generation/src/proof_generation/k/kore_convertion/language_semantics.py
```diff
@@ -330,10 +329,10 @@ def resolve_metavar(self, name: str) -> MetaVar:
             self._metavars[name] = MetaVar(name=len(self._metavars))
         return self._metavars[name]
 
-    def lookup_metavar(self, name: str) -> MetaVar:
-        if name in self._metavars:
-            return self._metavars[name]
-        raise KeyError(f'Variable name {name} not found in meta vars dict!')
```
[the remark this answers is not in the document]
MirceaS: I don't wanna keep around functions that are not used anywhere and are also untested. If we ever need it, it will be in the commit history.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -187,34 +187,52 @@ def current_configuration(self) -> Pattern:
         """Returns the current configuration."""
         return self._curr_config
 
+    @staticmethod
+    def assert_functional_pattern(semantics: LanguageSemantics, pattern: Pattern) -> None:
+        sym, _ = kl.deconstruct_nary_application(pattern)
+        # TODO: Should we also check that the arguments are functional?
```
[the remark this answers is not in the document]
MirceaS: This will happen as part of a different PR

generation/src/tests/unit/test_kore_language_semantics.py
MirceaS: Do you remember what you meant by this comment? Move what exactly?

generation/src/proof_generation/k/execution_proof_generation.py
MirceaS: I'm not sure I get your point. If you can make a function static (that is, it does not depend on the state of a particular instance) then that's great. A function should only be given as little as it needs to do what it's supposed to do.

generation/src/proof_generation/proofs/substitution.py
```diff
@@ -53,13 +55,23 @@ def top_univgen(self) -> ProofThunk:
...
+        func_subst_axiom_instance = func_subst_axiom(univ_evar.name)
+        return self.modus_ponens(
+            self.prop.imp_match_l(
+                self.modus_ponens(
+                    self.prop.imp_match_l(self.load_axiom(func_subst_axiom_instance), func_p_pf.conc), func_p_pf
+                ),
+                q_pf.conc,
+            ),
+            q_pf,
+        )
```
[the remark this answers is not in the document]
MirceaS: Can you please elaborate on exactly what kind of test you mean?

proofs/substitution.pretty-claim
```diff
@@ -1,9 +1,14 @@
-Load (μ X0 . X0)[]=1
-Load (phi0 -> (μ X0 . X0)[])=2
+SVar 0
+Mu 0
+Save
+MetaVar 0
+Load (μ X0 . X0)[]=0
+Implies
+Save
```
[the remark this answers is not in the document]
nishantjr: Ah, got it. Make sense.

generation/src/proof_generation/proofs/substitution.py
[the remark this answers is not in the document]
nishantjr: Add a unit test that passes in concrete proofs/axioms for each of the input proof thnks.

generation/src/proof_generation/k/execution_proof_generation.py
nishantjr: Isn't this part of the `ExecutionProofExp` class? Please use `self.pretty(pattern)`

proofs/generated-from-k/single-rewrite/foo-a.single-rewrite.pretty-proof
```diff
@@ -10,8 +75,5754 @@ Symbol ksym_Lbl'-LT-'generatedCounter'-GT-'
...
+Prop1
+Instantiate 0, 1
+ModusPonens
+Load Axiom ((phi0 -> (μ X0 . X0)[])[0: phi0] -> phi1)[0: phi0, 1: phi1][0: (phi0 -> (μ X0 . X0)[])[0: (phi0 -> (phi0 -> (μ X0 . X0)[])[0: phi1])][0: (phi0 -> (μ X0 . X0)[])[0: phi1], 1: (inhabitant · phi0)[0: phi0]][0: phi0, 1: phi1], 1: phi2][0: phi0, 1: phi1, 2: (kore_next · phi0)[0: phi2]][0: ksort_SortGeneratedTopCell, 1: ((ksym_Lbl'-LT-'generatedTop'-GT-' · phi0) · phi1)[0: (ksym_Lbl'-LT-'k'-GT-' · phi0)[0: ((kore_kseq · phi0) · phi1)[0: (((ksym_inj · phi0) · phi1) · phi2)[0: ksort_SortFoo, 1: ksort_SortKItem, 2: ksym_LblFooA'LParRParUnds'SINGLE-REWRITE-SYNTAX'Unds'Foo[]], 1: x0]], 1: x1], 2: ((ksym_Lbl'-LT-'generatedTop'-GT-' · phi0) · phi1)[0: (ksym_Lbl'-LT-'k'-GT-' · phi0)[0: ((kore_kseq · phi0) · phi1)[0: (((ksym_inj · phi0) · phi1) · phi2)[0: ksort_SortFoo, 1: ksort_SortKItem, 2: ksym_LblFooB'LParRParUnds'SINGLE-REWRITE-SYNTAX'Unds'Foo[]], 1: x0]], 1: x1]]=14
+ModusPonens
+Generalization 1
+ModusPonens
+ModusPonens
+Generalization 0
+ModusPonens
```
nishantjr, line 5827: @MirceaS This is a lot more than I expected to be in this proof.  The size of the binary proof has changed from 35 bytes to 12K. Any idea why there's so much more? Could you point me to all the changes that modify the main proof-generation.

generation/src/proof_generation/proofs/substitution.py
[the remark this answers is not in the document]
fiedlr: I think we already have a test for this?

generation/src/tests/unit/test_kore_language_semantics.py
fiedlr: I agree, but since this is burning, let's do it in a follow-up. Could you guys pair up on this task together on Monday?

generation/src/tests/unit/test_kore_language_semantics.py
```diff
@@ -651,38 +650,39 @@ def test_collect_functional_axioms() -> None:
...
-            b_symbol = mod.symbol('c', sort, input_sorts=(), is_functional=True, is_ctor=True)
-            c_symbol = mod.symbol('d', sort, input_sorts=(sort,), is_functional=True, is_ctor=True)
+            b_symbol = mod.symbol('b', sort, input_sorts=(), is_functional=True, is_ctor=True)
+            c_symbol = mod.symbol('c', sort, input_sorts=(sort,), is_functional=True, is_ctor=True)
+            d_symbol = mod.symbol('d', sort, input_sorts=(sort, sort), is_functional=True, is_ctor=True)
             a = a_symbol.app
             b = b_symbol.aml_symbol
             c = c_symbol.app
+            d = d_symbol.app
+            single_evar_krule = mod.rewrite_rule(kore_rewrites(sort.aml_symbol, EVar(0), EVar(0)))
```
fiedlr, line 660: What is this about? You cannot rewrite X to X without a symbol in Kore, can you?

generation/src/proof_generation/k/kore_convertion/language_semantics.py
```diff
@@ -692,9 +688,7 @@ def _convert_pattern(self, scope: ConvertionScope, pattern: kore.Pattern) -> Pat
                 exists_outer_sort_pattern: Pattern = self._convert_sort(scope, sort)
                 exists_var_pattern: Pattern = self._convert_pattern(scope, var)
                 exists_pattern: Pattern = self._convert_pattern(scope, pattern)
-                # TODO: This should be fixed after functional substitution is implemented
-                # assert isinstance(exists_var_pattern, EVar | SVar)
-                assert isinstance(exists_var_pattern, MetaVar)
+                assert isinstance(exists_var_pattern, EVar | SVar)
```
fiedlr, line 691:
```suggestion
                assert isinstance(exists_var_pattern, EVar)
```

there is no existential binder for set variables in matching logic

generation/src/proof_generation/k/execution_proof_generation.py
IljaZakharov: Agree with Nishant. It is not the same, you need to call self.pretty(self.current_configuration) and the rest similarly to keep notations pretty-printed.

generation/src/proof_generation/k/execution_proof_generation.py
IljaZakharov: There is a function stub `prove_substitutions` in the `SimplificationPerformer` above. It should do the same but for equational rules.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -376,34 +376,52 @@ def current_configuration(self) -> Pattern:
...
-        self.add_assumptions_for_rewrite_step(rule, substitution)
-
-        # Add the claim
-        self.add_claim(instantiated_axiom)
+        # Add the axioms/assumptions
+        self.add_axiom(rule.pattern)
+        for name, plug in substitution.items():
+            self.assert_functional_pattern(self.language_semantics, plug)
+            self.add_assumption(functional(0)(plug))
+            # TODO: Get rid of this hack:
```
fiedlr, line 407: It's not a hack. Just a matter of whether doing it systematically from a single axiom will be cheaper

fiedlr (approved):
Looks good, Mircea. Please just fix the remaining comments.

generation/src/tests/unit/test_kore_language_semantics.py
MirceaS: This test is not meant to recreate rules that we see in practice. It is a test that ensures that appropriate functional axioms are created for a rewrite event.

generation/src/proof_generation/k/execution_proof_generation.py
fiedlr: I'll fix the `prove_subst` today

generation/src/proof_generation/k/execution_proof_generation.py
MirceaS: I'm not sure I understand what you mean.
It is a hack since we're effectively adding as axioms multiple versions of a lemma which should be provable. Eventually we want to be able to, if not fully prove it, at least find a single form of the statement that we can then use to derive all the variations of it via alpha renaming.

generation/src/proof_generation/k/kore_convertion/language_semantics.py
MirceaS: You're right. I just went by the comment.

generation/src/proof_generation/proofs/substitution.py
```diff
@@ -9,25 +9,28 @@
...
 
 def forall(var: int) -> Notation:
     return Notation(f'forall_{var}', 1, neg(Exists(var, neg(phi0))), f'(∀ x{var} . {{0}})')
 
 
-func_subst_axiom = Implies(
-    Exists(0, equals(MetaVar(0, e_fresh=(EVar(0),)), EVar(0))),
-    Implies(forall(1)(phi1), phi1.apply_esubst(1, MetaVar(0, e_fresh=(EVar(0),)))),
-)
+def func_subst_axiom(forall_evar_name: int) -> Pattern:
```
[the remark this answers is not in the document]
MirceaS: We should really find a more systematic way to deal with notations that have variables in them. Having the definition take the bound variable as an input feels like we're going in the wrong direction. If that wasn't a thing then this wouldn't need to be a Notation. But for now, I suppose it would make things more efficient.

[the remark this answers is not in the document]
MirceaS: Actually, now that I think about how it's used I'm not so sure anymore that it should be a Notation. Its metavars are never instantiated. Its versions are just directly added as axioms, and then the metavars are instantiated when loading that axioms. So it never really acts as a Notation.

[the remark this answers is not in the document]
fiedlr: Ok, let's keep it like it is now.

PR #582, opened 2023-12-07
Adding `Lurk Tests` Job in Check.yml

Robertorosmaninho:
Closes #558 
Creating CI Tests for Lurk.


F-WRunTime:
For memory issues, queue the test on the [self-hosted, linux, prover] runners this has access to more memory. 

To run a job on a schedule you can run essentially a workflow as a cron. 
```
on: 
  schedule: 
    - cron: 'min hour DoM mon DoW' 
```
Cron expressions are tedious but you can get them pretty quick if you google cron expression generator 
But simply if you want to do "EveryDay at UTC 00:00:00" the expression is "0 0 * * *" -- This is probably a relatively reasonable time to run it for daily jobs to run during a lower usage window.

F-WRunTime (approved):

PR #585, opened 2023-12-07
Compose instantiation optimizer with memoization when optimizing proofs

malturki:
~~This PR depends on #584.~~

Update: The PR is ready for review.

This PR enables composing transformer interpreters and updates the optimization logic so that both instantiation optimization and memorization are applied. It also updates the proof snapshots.

Notes:
1. This PR is best reviewed by going over the final diff (rather than the individual commits)
2. Please go over the changes in the pretty-printed proofs to make sure they make sense


fiedlr (approved):

fiedlr:
Looks great, Musab! Thank you.

PR #591, opened 2023-12-07
Split Performance Numbers for Lurk

Robertorosmaninho:
Splitting GPU time in: Execution +Prove | Verify | Prove + Verify

Unfortunately, afaik is impossible to split between execution time and prove time in Lurk.
Due to the known issue of not being able to execute Lurk only in the CPU when GPU is available, we can't update the numbers regarding CPU execution using the same machine.


yzhang90 (dismissed):
LGTM

yzhang90 (approved):
LGTM

Robertorosmaninho:
@yzhang90 could you please add a new column with the total time? (Prove+verify)

yzhang90:
> @yzhang90 could you please add a new column with the total time? (Prove+verify)

done

PR #596, opened 2023-12-08
Notation Rework

MirceaS:
A rework of the way we handle `Notation`s that does not rely on passing notations into `ProofExp`s or maintaining a notation lookup in `PrettyOptions`. This makes the whole system more efficient (since we have no more lookup in a `Pattern` indexed map) and less error-prone (e.g. forgetting to add a newly defined notation into a list/worrying about notation duplication due to module imports).


MirceaS:
PR closed as we no longer wish to pursue this approach

PR #599, opened 2023-12-09
Memoizer: Remove unneeded logic for handling `None`

### LINKED ISSUE #580: `CountingInterpreter.finalize()` should return the empty set instead of None
Change the `CountingInterpreter.finalize()` to generate an empty set to avoid producing None and set `patterns_for_memoization` default to set() in the memoizer class.

malturki:
This closes #580


nishantjr (approved):
Great

PR #607, opened 2023-12-11
Fix xsubst functions for Instantiate in order to preserve notations

MirceaS:


generation/src/proof_generation/pattern.py
```diff
@@ -521,16 +534,16 @@ def instantiate(self, delta: Mapping[int, Pattern]) -> Pattern:
         return Instantiate(self.pattern.instantiate(unshadowed_delta), instantiated_subst)
 
     def apply_esubst(self, evar_id: int, plug: Pattern) -> Pattern:
-        # TODO: For "complete" substitutions (where all free metavars are replaced),
-        # and the substituted EVar does not occur, we should preserve the Instantiate,
-        # and apply the esubst to self.inst.
-        return self.simplify().apply_esubst(evar_id, plug)
```
IljaZakharov, line 527: Should we call `apply_esubst` for the `self.pattern` as well or I am missing something?

generation/src/proof_generation/pattern.py
```diff
@@ -78,9 +78,12 @@ def match(equations: list[tuple[Pattern, Pattern]]) -> dict[int, Pattern] | None
 
 
 class Pattern:
-    def evar_is_fresh(self, name: int) -> bool:
+    def evar_is_fresh_ignoring_metavars(self, name: int, ignored_metavars: frozenset[int]) -> bool:
         raise NotImplementedError
```
IljaZakharov, line 82: I would just add a new parameter with the default value, but it is up to you:

```
evar_is_fresh(self, name: int, ignoring_metavars: frozenset[int] = frozenset()) -> bool:
    ... 
```

IljaZakharov (commented):
I would just slightly change the function signature, but it looks good to me overall!

fiedlr (commented):
There's a bug in `evar_is_fresh` of `Instantiate` but this is not a solution.

We cannot ignore metavars from the notation in the definition of `evar_is_fresh`. The point is that such a notation can be instantiated where evar is not fresh, because there is no corresponding constraint, so `evar_is_fresh` has to return false.

We have to check `evar_is_fresh` on the simplification of the Instantiate.

fiedlr (changes requested):
Good catch, @MirceaS ! 

1. Please fix evar_is_fresh in Instantiate like this:
```
return self.pattern.simplify().evar_is_fresh(name)
```
2. Your solution to the notation issue with ESubsts/SSubst in `Instantiate` almost works, but let's not introduce that ignoring metavars function. Let's apply `Esubst/SSubst` to all the vals in `inst`, and apply them on the remaining metavars in `pattern` that are not instantiated in `inst`. This will work because:
a. if this `Instantiate` is instantiating a notation, then it will be fully instantiated so there are no remaining vars in pattern and it will get recognized
b. if it is not a notation, we do not care about pretty printing

MirceaS:
> Good catch, @MirceaS ! 
> 
> 
> 
> 1. Please fix evar_is_fresh in Instantiate like this:
> 
> ```
> 
> return self.pattern.simplify().evar_is_fresh(name)
> 
> ```
> 
> 2. Your solution to the notation issue with ESubsts/SSubst in `Instantiate` almost works, but let's not introduce that ignoring metavars function. Let's apply `Esubst/SSubst` to all the vals in `inst`, and apply them on the remaining metavars in `pattern` that are not instantiated in `inst`. This will work because:
> 
> a. if this `Instantiate` is instantiating a notation, then it will be fully instantiated so there are no remaining vars in pattern and it will get recognized
> 
> b. if it is not a notation, we do not care about pretty printing

For 1. see my earlier comment for an explanation on why it is not correct.
For 2., if we apply anything to the pattern in an Instantiate, the notation breaks, this includes putting an ESubst around a Metavar.
Also, at the moment, Instantiate is not used for anything other than holding Notations, even if we don't know they're notations (because they come from the binary format) and I don't think it males sense to use this type for anything else.
In light of that I think we should be more strict about making sure that objects of this type have the right shape, that is, that all metavars in the base pattern are instantiated and that there are no free s/eVars in the base pattern. These are very sensible things to check since you'll never instantiate only part of a notation and it doesn't make sense for a notation to have a certain variable free (though this point maybe we can revisit later if need be).
These checks are precisely what I introduce in this PR.

generation/src/proof_generation/pattern.py
MirceaS: The whole point of this PR is that we don't want to change the base pattern at all because it represents a Notation.
This is why I introduce these new checks to make sure that the base pattern is "clean" and doesn't need `apply_esubst` to be applied on it.

generation/src/proof_generation/pattern.py
MirceaS: That's a fair point but I think I'd rather keep them separate to point out that this not a normal option `evar_is_fresh` should have, it's here only for this particular use case, that users shouldn't really be concerned with.

fiedlr:
> > There's a bug in `evar_is_fresh` of `Instantiate` but the solution presented here is incorrect.
> > We cannot ignore metavars from the notation in the definition of `evar_is_fresh`. The point is that such a notation can be instantiated where evar is not fresh, because there is no corresponding constraint, so `evar_is_fresh` has to return false.
> 
> It isn't incorrect. The semantics of the new function I introduce `evar_is_fresh_ignoring_metavars` is not to check if an evar is fresh, it is to check if an evar is fresh, ignoring certain metavars in the process. I know full well that you can instantiate a constrain-less metavar with something where a particular evar is free but for the purpose of the check that we need to do in the substitution function, this ignoring semantics are what we need, because the metavars in question are being replaced.
> 
> > We have to check `evar_is_fresh` on the simplification of the Instantiate.
> 
> This is incorrect. As long as the substitution is not trivial, we will almost always have the varibale we're substituting for free in the base pattern. We want to be able to express something like substituting `sigma0` for `x0` in `phi0[0 -> x0]` to get `phi0[0 -> sigma0]`.

Mircea, I understand that you only keep out the check from metavariables in the `ignore` list. This is not the point. You are solving things at a completely different level where it needs to be, adding complications that can lead to serious issues in the future. You yourself admitted the function is ugly.

If you want `evar_is_fresh` to return True, you have to solve things at its root: replace the original metavariable in the notation, period. That is what I'm proposing. By replacing it in `inst`, you do not lose info about the notation and you can add a constraint. 

1. Let's say you have `phi0 /\ phi1`[0 |-> x]` and you want to apply [psi/x].

You can represent this as 

`phi0 /\ phi1[0 -> ESubst(phi0, x, psi), 1 -> ESubst(phi1, x, psi)]`

2. If you want to add a constraint, you can do so without any issues where the notation is used:

`phi0 /\ phi1[0 -> Metavar(0, e_fresh=EVar(0))]`

By performing the`evar_is_fresh` on the simplification, you do not lose anything because the function is side-effect free, it is only simplified within the function.

fiedlr:
For your implementation of `apply_ssubst` and `apply_esubst`, yes, this is fine. But the assert is not necessary (see examples above).

MirceaS:
I assume in your example you meant to say `phi0 /\ phi1[0 -> psi, 1 -> ESubst(phi1, x, psi)]`?
As for your other comment on performing `evar_is_fresh` on the simplification, consider my earlier example:
We have `phi0[0 -> x0]` and we want to substitute `x0` by `sigma0`. After performign the simplification, we get `x0`, but if we check that `x0` is fresh in `x0` we obviously get `false`, so that doesn't work.
What I need is to check that the variable is fresh only in the base pattern, and not in the instantiations.
If we don't check that we can end up with scenarios where the base pattern is `phi0 /\ x0` [0 -> phi1]`. If we don't do the check, we end up with no change when substituting `x0`.

fiedlr:
@MirceaS 

1. Yes, sorry, I edited the example. But with a wrapped `ESubst` so that this is systematic. In this special case, it is `psi`, you are correct, because we need to check `pattern == plug` to avoid the redundant substitution assertion error. I hope the general pattern is clear. 

2. The check is recursive. The issue you are mentioning is elsewhere, namely in combining instantiations and substitutions that result in a redundant ESubst/SSubst. 

This won't happen with the updated `apply_esubst` and `apply_ssubst`. You should not use the constructors directly in most cases,

phi0[0 -> x0].apply_esubst(x0, sigma0).evar_is_fresh(0)

~>

phi0[0 -> sigma0].evar_is_fresh(0)

~> (with simplify in implementation)

sigma0.evar_is_fresh(x0) = True


*Note*

- We should audit where we use `ESubst/SSubst` constructors directly. These constructors should be used directly only in very specific circumstances because the automatic reductions might not be applied.

- `evar_is_fresh` in `ESubst` assumes that everything is fully reduced. You see now why these assumptions are important for well-formedness?

MirceaS:
@fiedlr `ESubst(x, x, psi)` is impossible to express in our current format. `ESubst` can only have as the base pattern a metavar or another E/SSubst. Pending substitutions only exist for those 3 constructs.
The rest of your reply I don't really understand.
> `.apply_esubst(x0, sigma0).evar_is_fresh(0)`
There was never any talk of checking freshness on the result of an `apply_esubst`.

fiedlr:
> 1. we need to check `pattern == plug` to avoid the redundant substitution assertion error. I hope the general pattern is clear.

Yes, that's what I said

```
we need to check pattern == plug to avoid the redundant substitution assertion error. I hope the general pattern is clear.
```

The general pattern is wrapping in `ESubst`, in this special case we have to provide the argument directly in the `apply_esubst` function.

2. I mean you mentioned freshness in your own response, quoting:

> ... consider my earlier example: We have `phi0[0 -> x0]` and we want to substitute `x0` by `sigma0`. After performign the simplification, we get `x0`, but if we check that `x0` is fresh in `x0` we obviously get `false`, so that doesn't work...

I showed that it does work, that's all.

generation/src/proof_generation/pattern.py
nishantjr: I agree with @IljaZakharov . Isn't the functionality identical when `ignoring_metavars` is empty?

generation/src/proof_generation/pattern.py
```diff
@@ -521,16 +534,13 @@ def instantiate(self, delta: Mapping[int, Pattern]) -> Pattern:
         return Instantiate(self.pattern.instantiate(unshadowed_delta), instantiated_subst)
 
     def apply_esubst(self, evar_id: int, plug: Pattern) -> Pattern:
-        # TODO: For "complete" substitutions (where all free metavars are replaced),
-        # and the substituted EVar does not occur, we should preserve the Instantiate,
-        # and apply the esubst to self.inst.
-        return self.simplify().apply_esubst(evar_id, plug)
+        assert self.pattern.evar_is_fresh_ignoring_metavars(evar_id, frozenset(self.inst.keys()))
```
nishantjr, line 537: The freshness criteria is more important than completeness.
To be honest, I feel the completeness requirement may be too strong. Is freshness all we need?
I hadn't thought of this ignoring mv approarch when I wrote the comment.

Lets turn this into two cases. If the variable is fresh, and when it is not. If not we fallback to performing the simplification. 

There are lots of edge cases here. It needs more unit tests. I'd put it on test for each combination of completeness and freshness, including freshness when is concrete vs due to a metavar criteria, as well as one for nested intantiations.

nishantjr (changes requested):

generation/src/proof_generation/pattern.py
```diff
@@ -422,14 +429,16 @@ class ESubst(Pattern):
     var: EVar
     plug: Pattern
 
-    def evar_is_fresh(self, name: int) -> bool:
+    def evar_is_fresh_ignoring_metavars(self, name: int, ignored_metavars: frozenset[int]) -> bool:
         if self.var.name == name:
-            return self.plug.evar_is_fresh(name)
+            return self.plug.evar_is_fresh_ignoring_metavars(name, ignored_metavars)
 
         # We assume that at least one instance will be replaced
```
nishantjr, line 436:
```suggestion
        # This check may be slightly stronger than necessary in the case where `self.var` is fresh in `self.pattern`,
        # but not in `self.plug`. Since `evar_if_fresh` may return `False` when freshness in not known this is OK.
        # Indeed this option may be better for efficiency, since it is unlikely that such an `ESubst` is constructed.
```

nishantjr:
I like the approach Mircea, but as pointed out the definition for `Instantiate` isn't sound.
However, I think we can make it sound without losing notation. See comments above.

Before we merge this lets:

1. [ ] Fix the unsoundness
2. [ ] Merge `evar_is_fresh` and `evar_is_fresh_ignoring_mv`
3. [ ] Extend this to SVars as well.
4. [ ] Add unit tests for the various edge cases.

fiedlr:
> I like the approach Mircea, but as pointed out the definition for `Instantiate` isn't sound. However, I think we can make it sound without losing notation. See comments above.
> 
> Before we merge this lets:
> 
> 1. [ ]  Fix the unsoundness
> 2. [ ]  Merge `evar_is_fresh` and `evar_is_fresh_ignoring_mv`
> 3. [ ]  Extend this to SVars as well.
> 4. [ ]  Add unit tests for the various edge cases.

@MirceaS and I don't agree with 2., and 3. is not critical for now because we are not using SSubsts for functional substitutions. That is why we should skip 2. and 3. and just focus on 1. and 4., and get this merged asap because @MirceaS is leaving for the winter break tomorrow.

generation/src/tests/unit/test_pattern.py
```diff
@@ -58,10 +72,12 @@
...
         # Instantiate/Notation
         [
-            Instantiate(App(phi0, phi1), frozendict({0: phi2})),
+            Instantiate(App(phi0, phi1), frozendict({0: phi2, 1: phi1})),
             0,
             sigma1,
-            App(ESubst(phi2, EVar(0), sigma1), ESubst(phi1, EVar(0), sigma1)),
+            Instantiate(
+                App(phi0, phi1), frozendict({0: ESubst(phi2, EVar(0), sigma1), 1: ESubst(phi1, EVar(0), sigma1)})
+            ),
```
fiedlr, line 80: Please add more tests where the contained `ESubsts` are simplified

fiedlr (approved):
Looks good, @MirceaS , but honestly I would expect more tests. There is also no tests on the new `ignoring` functionality. We have to implement it this week in a follow-up.

generation/src/proof_generation/aml/syntax.py
```diff
@@ -407,14 +414,20 @@ def __post_init__(self) -> None:
...
-
-        # We assume that at least one instance will be replaced
-        return self.pattern.evar_is_fresh(name) and self.plug.evar_is_fresh(name)
-
-    def metavars(self) -> set[int]:
+            return self.plug.evar_is_fresh_ignoring_metavars(name, ignored_metavars)
+
+        # This check may be slightly stronger than necessary in the case
+        # where `self.var` is fresh in `self.pattern`, but `name` is not fresh in `self.plug`.
+        # Since `evar_if_fresh` may return `False` when freshness in not known this is OK.
```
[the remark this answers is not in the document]
MirceaS: This comment was here before I started writing this PR but I agree that. it is out of date now.
What do you mean by "incomplete" here though?

generation/src/tests/unit/test_pattern.py
MirceaS: Simplified how? I'm not sure I follow

fiedlr: For example, before:

`0 |-> x /\ sigma2`

after

`0 |-> sigma1 /\ sigma2`

Also we need tests for the ignoring component of `evar_is_fresh_ignoring_metavars`

generation/src/proof_generation/aml/syntax.py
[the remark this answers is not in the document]
fiedlr: `assert not self.evar_is_fresh(self.var)` does not guarantee `self.var` is free thanks to how constraints in metavars work.

For this assumption to be made that at least one instance of `plug` is plugged, we have to have a function `self.evar_is_free` and a respective `e_free` array in metavar constraints.

generation/src/tests/unit/test_pattern.py
MirceaS: This is not a feature we support at the moment, that is, propagating esubstitutions through ESubsts

PR #609, opened 2023-12-12
Implemementing the skeleton version of Cairo Checker

Robertorosmaninho:
This PR contains: 
- a preliminary version of `Pattern` and its constructors
- a preliminary version of `execute_instructions`
- more details of what this PR implements can be found at #610 TODO List

Some important things to notice in this implementation are:
1. We can only use match with `felt252` and `enums`
1. In `match` with `felt252`, "numbers must be sequential starting from 0", so we *must* fill the gap between 30 and 137.
2. in `match` with `enums`, "variants must be in the same order as enum declaration.", so we need to implement all `enum` cases.


cairo/src/verifier/pattern.cairo
```diff
@@ -0,0 +1,244 @@
...
+// Boxes.
+use box::{Box, BoxTrait};
+
+use core::traits::{Index, Default, Felt252DictValue};
+use core::array::Array;
+
+type Id = u8;
+type IdList = Array<Id>;
+
+fn copyList(list: @IdList) -> IdList {
```
yzhang90, line 20: Is the copy list reversed ?

yzhang90 (approved):
LGTM

PR #612, opened 2023-12-12
Implement prove_simplification method

### LINKED ISSUE #473: Implement `ExecutionProofExp.prove_simplification`
1. Assume axiom `p1 k= p2 -> phi0[p1/x] k= phi0[p2/x]`
2. Define a function that proves:

   ```
           p1 k= p2
   -----------------------
   phi0[p1/x] k= phi0[p2/x]
   where $x$ is an `app_ctx_hole` in `phi0`.
   ```
*Sketch for 2.* Load the axiom given by 1. and apply MP.

3. Implement a method `prove_simplification`:

    ```
    def prove_simplification(ordinal, substitution, location: tuple[int]) -> ProofThunk
    ```

    that proves that the subterm of `ProofGenExp.current_config` at `location`
    can be simplified using the rule with ordinal `ordinal`.
    That is, suppose:

    * `ProofGenExp.current_config` is `<k> four(3) </k>`.
    * `location` is `(0)`, and
    * `ordinal` refers to the rule `four(I) = 4`

    Then the function `prove_simplification` returns the proof of `<k> four(3) </k> k= <k> 4 </k>`

*Sketch for 3.*
The function will use `SimplificationPerformer::update_subterm` function to plug a hole (element variable) into `loc = (0)` of `current_config`, this will be `phi0`. Then it will instantiate the equational rule given by `ordinal` with the given `substitution` to get `p1 := four(3)` (lhs of the equational ruile) and `p2 = 4` (rhs of the equational rule), and apply 2. with this `phi0` and `p1, p2`.

IljaZakharov:
Closes #473


generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -55,12 +65,19 @@ def update_configuration(self, new_current_configuration: Pattern) -> None:
 
     def apply_simplification(
         self, ordinal: int, substitution: dict[int, Pattern], location: Location
-    ) -> SimplificationInfo:
+    ) -> SimplificationSummary:
         # Check that whether it is the first simplification or not
         if not self._simplification_stack:
             sub_pattern = self.get_subterm(location, self._curr_config)
+            config_with_hole = self.update_subterm(location, self._curr_config, self.HOLE)
```
fiedlr, line 72: I do not think we need to save this in a special data structure

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -219,16 +242,49 @@ def rewrite_event(self, rule: KRewritingRule, substitution: dict[int, Pattern])
...
             performer.apply_simplification(ordinal, substitution, location)
-            # TODO: Do some proving here ...
+            # TODO: Blocked by #601, #462
+            # simplification_conf_proof = self.prove_simplification(simplification_summary, substitution)
 
         # Update the current configuration
         self._curr_config = self._simplification_performer.simplified_configuration
 
+    def prove_simplification(
+        self, summary: SimplificationSummary, _tmp_metavar_substitutions: dict[int, Pattern]
```
fiedlr, line 252: What are these _tmp_metavar_substitutions?

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -30,12 +30,22 @@
 @dataclass
 class SimplificationInfo:
     location: Location
+    configuration_with_hole: Pattern  # Relative to previous in stack
     initial_pattern: Pattern  # Relative to previous in stack
     simplification_result: Pattern
     simplifications_remaining: int
```
fiedlr, line 36: Please add `proof: ProofThunk` here

Issue #614, opened 2023-12-12
Fix the transfer5000 example


fiedlr:
Description of the updates is in this thread: https://runtimeverification.slack.com/archives/C05MF3KEWNM/p1702368948400919

1. Update the direct implementation `transfer5000` in every one of our zkVMs: (#615 and #619)
2. Update the pi2 implementation in [ `https://github.com/fiedlr/k-proof-generation/tree/main/examples/csl23` ](https://github.com/fiedlr/k-proof-generation/blob/main/examples/csl23/blockchain/transfer-batch.imp5) (this was not yet transferred over to `proof-checker` repo because we are not supporting it yet)
3. Regenerate the MM slice and update the translation of 2. [consult @malturki  on this)
4. Rerun benchmarks on both direct implementations and the pi2 implementation

PR #615, opened 2023-12-12
Update risc0 transfer5000 direct implementation and update performance numbers

Robertorosmaninho:
Part of #614

- Update transfer and transfer5000 Risc0 CSL direct implementations
- Fix a bug when counting verification time
- Reorganize benchmark order on `bench.sh`
- Update `Performance.md` with the modifications described above and set the Risc0 version used to run the experiments.


yzhang90 (approved):
LGTM

PR #616, opened 2023-12-12
Remove sorts from kore-and and kore-or

fiedlr:
Fixes pattern matching over `kore_and` and `kore_or` by removing the sort argument that is not used in the notation definition, as these do not contain the sort `phi0` argument in their definition, so pattern match fails. 

See https://runtimeverification.slack.com/archives/C05MF3KEWNM/p1702398036443069 for more explanation

*Code review*

Please check out this branch and make sure that the sort information was removed from all the applications of `kore_and` and `kore_or`.


malturki (approved):
It seems you've got them all.

IljaZakharov (approved):
Looks good! I would merge it now as I need these fixes to apply `assert_matches`.

PR #617, opened 2023-12-13
Implement proof generation for `SimplificationPerformer`

fiedlr:
This PR extends `SimplificationPerformer` with `SimplificationProver` that uses the info on the simplification stack to produce an equality proof of `input_term k= fully_simplified_input_term`.


generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -33,13 +33,15 @@ class SimplificationInfo:
...
     simplifications_remaining: int
+    proof: ProofThunk
 
 
 class SimplificationPerformer:
     def __init__(self, language_semantics: LanguageSemantics, init_config: Pattern):
         self._language_semantics = language_semantics
         self._curr_config = init_config
         self._simplification_stack: list[SimplificationInfo] = []
+        self.proof = None # If None, the batch has not been proved yet
```
nishantjr, line 44: Perhaps we can initialize this with  a proof for `init_config = curr_config`?
That removes a special case.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -75,13 +91,21 @@ def apply_simplification(
         simplifications_ramaining = self._language_semantics.count_simplifications(simplified_rhs)
 
         # Create the new info object and put it on top of the stack
-        new_info = SimplificationInfo(location, sub_pattern, simplified_rhs, simplifications_ramaining)
+        new_info = SimplificationInfo(location, sub_pattern, simplified_rhs, simplifications_ramaining, proof_given_by_ordinal(ordinal, substitution))
         self._simplification_stack.append(new_info)
```
nishantjr, line 95: My feeling is that a new `SimplificationInfo` needs to be created by `enter_context`.
We could combine them for now, since every simplification creates a context. We'd need to split them up again
when hooks are implemented.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -218,11 +244,15 @@ def rewrite_event(self, rule: KRewritingRule, substitution: dict[int, Pattern])
...
     def simplification_event(self, ordinal: int, substitution: dict[int, Pattern], location: Location) -> None:
         with self._simplification_performer as performer:
+            performer.enter_context(location)
             performer.apply_simplification(ordinal, substitution, location)
-            # TODO: Do some proving here ...
 
         # Update the current configuration
-        self._curr_config = self._simplification_performer.simplified_configuration
+        # This completes rewrite step 1 after all simplifications
+        if self._simplification_performer.proof is not None: # This means that we finished the batch and proof is ready
```
IljaZakharov, line 252: I would stick to some property flag like `in_simplification` or something

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -218,11 +244,15 @@ def rewrite_event(self, rule: KRewritingRule, substitution: dict[int, Pattern])
 
     def simplification_event(self, ordinal: int, substitution: dict[int, Pattern], location: Location) -> None:
         with self._simplification_performer as performer:
+            performer.enter_context(location)
             performer.apply_simplification(ordinal, substitution, location)
-            # TODO: Do some proving here ...
```
IljaZakharov, line 222: I propose to load the simplification proof here instead like this:

```
conf_with_hole, rule, substitutions = performer.enter_context(location)
one_step_proof = self.apply_simplification(conf_with_hole, rule, substitutions)
perfromer.load(one_step_proof)
```

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -53,18 +55,32 @@ def update_configuration(self, new_current_configuration: Pattern) -> None:
...
         # Check that whether it is the first simplification or not
         if not self._simplification_stack:
             sub_pattern = self.get_subterm(location, self._curr_config)
         else:
             sub_pattern = self.get_subterm(location, self._simplification_stack[-1].simplification_result)
 
-        # Get the rule and remove extra variables added by K in addition to the ones in the K file
+        # Create the new info object and put it on top of the stack
+        # TODO: simplifications_remaining=0 here will probably not work with __exit__ as intended
+        new_info = SimplificationInfo(location, sub_pattern, sub_pattern, 0, eq_id) # ADD phi = phi to kore lemmas called as eq_id and instantiate it with phi := init_config/curr_config at this stage
```
IljaZakharov, line 70: Apparently, the code in `__exit__` should be adjusted if we add an unexpected element.

IljaZakharov (commented):
I like the structure overall! But still, I explained yesterday that I would prefer _loading_ proofs to the perfromer instead of doing it in the class. Shortly, I can summarize my reasoning as this:

- We don't need the location itself and the stack for proving. Let's keep it in the performer. But we need the configuration with the hole and side substitutions to infer the proof for a single proof rule. That is why I would just return necessary data to the proof expression instead of working with the simplification info augmented by the proof and then just load the proof before exiting the `with` block. Check out my draft PR for the #473 that shows what we do really need for the proof.
-  Maybe I am wrong, please correct me. But it is not possible to prepare the proof outside of a proof expression: we need notations, axioms, imported proof expressions literally. But I would avoid inheriting from the proof expression, it is better just to merge classes then if we can't keep separation of concerns.
- Adding proving to functions that process simplification makes tests much more complicated. I would try to test processing hints separately from preparing proofs. Otherwise, we simply cannot run existing tests until we finish the proving functionality.

So, my proposal is to keep everything from this PR but extract proving functions to the proof expressions.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -53,18 +55,32 @@ def update_configuration(self, new_current_configuration: Pattern) -> None:
...
+
+        return new_info
+
+    def apply_simplification(
+        self, ordinal: int, substitution: dict[int, Pattern], location: Location
+    ) -> None:
+        # Get the rule and assert the lhs equals current config that we are simplifying
         rule = self._language_semantics.get_axiom(ordinal)
         assert isinstance(rule, KEquationalRule), 'Simplification rule is not equational'
+        assert rule.left == self._curr_config
```
IljaZakharov, line 81: The rule is not instantuated/has substitutions pending

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -75,13 +91,21 @@ def apply_simplification(
         simplifications_ramaining = self._language_semantics.count_simplifications(simplified_rhs)
 
         # Create the new info object and put it on top of the stack
-        new_info = SimplificationInfo(location, sub_pattern, simplified_rhs, simplifications_ramaining)
+        new_info = SimplificationInfo(location, sub_pattern, simplified_rhs, simplifications_ramaining, proof_given_by_ordinal(ordinal, substitution))
         self._simplification_stack.append(new_info)
 
-        return new_info
+        # Update current config
+        self._curr_config = simplified_rhs
```
IljaZakharov, line 98: This is not the configuration! This is just the simplification of a function call. But it is doable to get the configuration there.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -75,13 +91,21 @@ def apply_simplification(
...
 
-    def __enter__(self) -> SimplificationPerformer:
-        return self
+    def proof_given_by_ordinal(self, ordinal, substitution) -> ProofThunk:
```
IljaZakharov, line 100: Should it be the equation for configurations or for the applied rule? There is no pattern argument that should be used for proving out the equation from.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -75,13 +91,21 @@ def apply_simplification(
...
+        # return basically an instantiated equality axiom used in apply_simplification as ProofThunk
+        pass
+
+    def update_equality_proof(self, proof: ProofThunk) -> ProofThunk:
```
IljaZakharov, line 104: This is the proof for configurations then.

fiedlr, line 104:
```suggestion
    def apply_framing_lemma(self, equality_proof: ProofThunk, context: Pattern) -> ProofThunk:
```

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -98,11 +122,13 @@ def __exit__(
                 last_info.simplification_result = self.update_subterm(
                     exhausted_info.location, last_info.simplification_result, exhausted_info.simplification_result
                 )
+                last_info.proof = update_equality_proof(last_info.proof)
```
fiedlr, line 125:
```suggestion
                last_info.proof = equality_transitivty(last_info.proof, apply_framing_lemma(exhausted_info.proof, last_info.make_context(exhausted_info.location)))
```

fiedlr: Please rename `last_info` to `parent_node`
Please rename `exhausted_info` to `child_node`

generation/src/proof_generation/k/execution_proof_generation.py
IljaZakharov: Decided to go with the separate class instead of loading.

cairo/Scarb.toml
```diff
@@ -1,6 +1,7 @@
 [package]
 name = "ml_checker_cairo"
 version = "0.1.0"
+edition = "2023_01"
```
IljaZakharov, line 4:
```suggestion
```

fiedlr (commented):
Looks great, let's merge.

IljaZakharov (approved):
LGTM

PR #619, opened 2023-12-13
Fix ZKLLVM and Lurk examples

yzhang90:
Part of #614


Performance.md
```diff
@@ -12,10 +12,10 @@
...
 |     Examples     | CPU Circuit Gen Time | CPU Prove+Verify Time | GPU Time |
 |:----------------:|:--------------------:|:---------------------:|:--------:|
-| perceptron       |                 0.95 |                 0.135 |          |
-| svm5             |                 0.96 |                 0.135 |          |
-| transfer         |                 0.97 |                 0.133 |          |
-| transfer-batch   |                0.797 |                 40.34 |          |
+| perceptron       |                 0.45 |                 0.650 |          |
+| svm5             |                 0.45 |                 0.630 |          |
+| transfer         |                 0.44 |                 0.630 |          |
+| transfer5000     |                0.823 |                38.842 |          |
```
Robertorosmaninho, line 18: Let's keep 3 places after the decimal point as a pattern

Performance.md
```diff
@@ -24,14 +24,14 @@
 | perceptron       |    11   |            |           0.84 |           0.597 |          1.437 |
 | svm5             |    9    |            |           0.82 |           0.607 |          1.427 |
 | transfer         |    34   |            |          2.205 |           0.595 |          2.800 |
-| transfer5000*    |  330387 |            |        487.322 |           4.899 |        492.221 |
+| transfer5000*    |  505037 |            |              ∞ |               ∞ |              ∞ |
```
Robertorosmaninho, line 27: Same thing about keeping 3 places after the decimal point as a pattern

Robertorosmaninho (commented):
A few comments about the way to display the numbers. Also, if we're not displaying the numbers for the CPU execution of Lurk CSL direct implementations due to current Lurk issues, we shouldn't display these numbers for the Pi2 implementations as well to be consistent.

Robertorosmaninho (changes requested):
See my comments above

Performance.md
yzhang90: done

Performance.md
yzhang90: done

Robertorosmaninho:
Just the last comment:
> ... Also, if we're not displaying the numbers for the CPU execution of Lurk CSL direct implementations due to current Lurk issues, we shouldn't display these numbers for the Pi2 implementations as well to be consistent.

Robertorosmaninho (approved):

yzhang90:
> Just the last comment, @yzhang90:
> 
> > ... Also, if we're not displaying the numbers for the CPU execution of Lurk CSL direct implementations due to current Lurk issues, we shouldn't display these numbers for the Pi2 implementations as well to be consistent.

done

PR #620, opened 2023-12-13
Implement function for proving the equation of of a given equality rule

### LINKED ISSUE #601: Removing clutter introduced by `equational_as` to use equalities in proofs
# Intro

Equational rules are usually in the form of `equational_as` like this (*):

```
Evar(0) = foo(x) -> Evar(0) = bar(x) /\ Top  //  (foo(x) = bar(x))
```

We use the `kore_in` notation to handle saving these.

But on the proof level, we should get rid of the unnecessary clutter. This issue proposes how.

# Solution

`SimplificationPerformer` can apply `base_substitutions` given by the `requires` clause to get from (*) to something like

```
 foo(x) = foo(x) -> foo(x) = bar(x) /\ Top  
```

We need to get rid of the lhs and the additional top on the rhs. For now, we will do so by adding some extra axioms.

1. To get rid of the lhs, add an axiom stating this:

```
(phi0 = phi0 -> psi) -> psi
```

2. To get rid of the rhs, we have:
```
(phi0 /\ Top) -> phi0
```

We'll prove these lemmas properly later.

TODO: Try to think of a lemma that uses `equational_as` notation directly.

IljaZakharov:
- Removes tops from the requirement clause
- Removes the requirement itself
- Removes the ensures

Uses several axioms added to the `kore` proof expression. These axioms should be proved later.

Closes #601


generation/src/proof_generation/proofs/kore.py
```diff
@@ -232,6 +273,101 @@ def equality_with_subst(self, phi: Pattern, equality: ProofThunk):
...
+
+    def reduce_left_top_in_imp(self, phi: ProofThunk):
+        """
+                k⊤ k⋀ phi1 k-> phi2
+        ---------------------------
+                phi1 k-> phi2
+        """
+        sort, left_conj, conclusion = kore_implies.assert_matches(phi.conc)
+        _, right = kore_and.assert_matches(left_conj)
+        return self.modus_ponens(
```
fiedlr, line 348: For all the lemmas that remove top, I think we can do this more systematically with these four:

```
T /\ phi = phi
phi /\ T = phi
phi = psi -> psi = phi
psi1 = psi2   phi1[psi1/x] 
------------------------------------
phi1[psi2/x]
for any variable `x`
```
The last one is a stronger version of the one we have where `x` needs to be a hole. Actually, `x` does not need to be a app ctx hole, it can be a hole in any context `phi1` (see Matching mu-Logic TR), which is equivalent with the last lemma here:

(=>) if there are n occurrences of phi1 in x, apply the context lemma n times for each occurrence, i.e., the considered context will have a whole xi for i-th occurrence of x
(<=) pick phi1 to be a context where x is a hole

fiedlr: This systematic approach is better because it will allow us to remove tops from conjunctions appearing anywhere, including n-ary conjunctions (nested conjunctions)

generation/src/proof_generation/proofs/kore.py
```diff
@@ -232,6 +273,101 @@ def equality_with_subst(self, phi: Pattern, equality: ProofThunk):
             equality,
         )
 
+    def reduce_equational_as(self, phi: ProofThunk):
```
fiedlr, line 276: We can get also rid of this by having the lemmas I presented in the other comment plus

```
phi2 /\ phi2 = phi2
```

generation/src/tests/unit/test_execution_proof_generation.py
```diff
@@ -396,6 +398,54 @@ def test_subpattern_batch():
...
+    a_symbol = semantics.get_symbol('a')
+    b_symbol = semantics.get_symbol('b')
+    node_symbol = semantics.get_symbol('node')
+    reverse_symbol = semantics.get_symbol('reverse')
+    tree_sort = semantics.get_sort('SortTree').aml_symbol
+
+    # Create a new proof expression
+    proof_expr = ExecutionProofExp(semantics, init_config=top())
+
+    # reverse(a) <-> a
```
fiedlr, line 412:
```suggestion
    # reverse(a) = a
```

generation/src/tests/unit/test_execution_proof_generation.py
```diff
@@ -396,6 +398,54 @@ def test_subpattern_batch():
...
+    base_case_a = semantics.get_axiom(2)
+    assert isinstance(base_case_a, KEquationalRule)
+    rule_with_substitution = base_case_a.pattern.apply_esubsts({0: a_symbol.app(), 1: a_symbol.app()})
+
+    rule_proof_thunk = make_pt(rule_with_substitution)
+    expected_equation = kore_equals(tree_sort, tree_sort, reverse_symbol.app(a_symbol.app()), a_symbol.app())
+    equation_proof = proof_expr.prove_equality_from_rule(rule_proof_thunk)
+    assert equation_proof(BasicInterpreter(phase=ExecutionPhase.Proof)).conclusion == expected_equation
+
+    # reverse(b) <-> b
```
fiedlr, line 422:
```suggestion
    # reverse(b) = b
```

generation/src/tests/unit/test_execution_proof_generation.py
```diff
@@ -396,6 +398,54 @@ def test_subpattern_batch():
...
+    base_case_b = semantics.get_axiom(3)
+    assert isinstance(base_case_b, KEquationalRule)
+    rule_with_substitution = base_case_b.pattern.apply_esubsts({0: b_symbol.app(), 1: b_symbol.app()})
+
+    rule_proof_thunk = make_pt(rule_with_substitution)
+    expected_equation = kore_equals(tree_sort, tree_sort, reverse_symbol.app(b_symbol.app()), b_symbol.app())
+    equation_proof = proof_expr.prove_equality_from_rule(rule_proof_thunk)
+    assert equation_proof(BasicInterpreter(phase=ExecutionPhase.Proof)).conclusion == expected_equation
+
+    # reverse(node(T1, T2)) <-> node(reverse(T2), reverse(T1))
```
fiedlr, line 432:
```suggestion
    # reverse(node(T1, T2)) = node(reverse(T2), reverse(T1))
```

fiedlr (approved):
Great work, Ilja. I think we can have a more systematic approach but this is not your fault, as the task itself gives an adhoc solution.

Let's not block this any further, because the code looks good otherwise.
Please refrain from adding more lemmas of this kind. If you come across this need, we need to do the systematic approach. I will add a task for refactoring `prove_equality_from_rule` based on the systematic approach.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -224,6 +224,42 @@ def simplification_event(self, ordinal: int, substitution: dict[int, Pattern], l
...
+                if match := kl.kore_top.matches(left):
+                    proof = self.kore_lemmas.reduce_left_top_in_imp(exp)
+                    rest_requires = right
+                    return reduce_requirement_rec(proof, rest_requires)
+                elif match := kl.kore_top.matches(right):
+                    proof = self.kore_lemmas.reduce_right_top_in_imp(exp)
+                    rest_requires = left
+                    return reduce_requirement_rec(proof, rest_requires)
+                else:
+                    return exp, cached_requires
```
fiedlr, line 251: I think here we need to call `reduce_requirement_rec` too, right?

The top can be also in the middle:

phi1 /\ phi2 /\ top /\ phi3 /\ phi4

This would not catch this

generation/src/proof_generation/proofs/kore.py
```diff
@@ -210,11 +210,52 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
...
+remove_top_imp_right_axiom = Implies(
+    kore_implies(phi0, kore_and(phi1, kore_top(phi0)), phi2), kore_implies(phi0, phi1, phi2)
+)
+
+# TODO: Requires a proof
+# (phi2:{phi0} k⊆ (phi2 k/\ phi2):{phi0}):{phi1} k-> phi3):{phi1} -> phi3
+reduce_equational_as_axiom = Implies(kore_implies(phi1, kore_equational_as(phi0, phi1, phi2, phi2, phi2), phi3), phi3)
+
+# TODO: Requires a proof
+# (phi2:{phi0} k⊆ phi2:{phi0} k-> phi3):{phi1} -> phi3
```
fiedlr, line 236: Here the systematic approach would be to add

`phi <= phi = Top`

generation/src/proof_generation/proofs/kore.py
```diff
@@ -210,11 +210,52 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
...
+)
+
+# TODO: Requires a proof
+# ((phi1:{phi0} k/\ k⊤:{phi0}) k-> phi2):{phi0} -> (phi1:{phi0} k-> phi2)):{phi0}
+remove_top_imp_right_axiom = Implies(
+    kore_implies(phi0, kore_and(phi1, kore_top(phi0)), phi2), kore_implies(phi0, phi1, phi2)
+)
+
+# TODO: Requires a proof
+# (phi2:{phi0} k⊆ (phi2 k/\ phi2):{phi0}):{phi1} k-> phi3):{phi1} -> phi3
```
fiedlr, line 232: Here you would first do phi2 /\ phi2 = phi2, and then use phi2 <= phi2 as top, and then use the top lemmas to remove it

PR #623, opened 2023-12-13
Fix dict order relevance bug

MirceaS:


IljaZakharov (approved):
Good Catch! I also played with dictionaries and without sorting they do preserve the original order, so we need sorting using keys.

PR #630, opened 2023-12-14
Drop unused sorts from `kore_next` and `kore_bottom` notations

malturki:
This PR is a continuation of #616 where we remove sorts from notations that don't use them to address an issue with matching. Here we remove sorts from `kore_next` and `kore_bottom`.

Note: The first commit addresses a minor formatting error that I was facing locally, where the checker was complaining about using double-quoted strings.


nishantjr (approved):
LGTM

PR #631, opened 2023-12-14
Preparing Cairo Checker to implement `execute_instructions`

Robertorosmaninho:
This PR:
- Sets the scarb's version to Nightly
- Implements `PartialEq` for Pattern
  - Enable the `asserts` that couldn't be checked before on the unit tests 
- Fixes `Clone` for Pattern
- Makes `Stack` parametric and :
   - Implements `StackTrait` and its `StackTraitImpl`
   - Implements `last` as a lookup for `StackTrait`
   - Implements `ClaimTrait` and its `ClaimTraitImpl`
- Delete the remaining mentions to `stack_size`
- Delete `Entry` as we can use `Term` in the same way


cairo/src/stack.cairo
```diff
@@ -55,16 +57,89 @@ impl StackTraitImpl of StackTrait {
...
+            match term {
+                Option::Some(term) => {
+                    if i == self.len - 1 {
+                        pop_term = term;
+                        break;
+                    }
+                    new_stack.append(term);
+                    i += 1;
+                },
+                Option::None => { break; }
```
yzhang90, line 123: Shall we panic here?

cairo/src/stack.cairo
```diff
@@ -55,16 +57,89 @@ impl StackTraitImpl of StackTrait {
...
+        if self.is_empty() {
+            panic!("Insufficient stack items.");
+        }
+
+        let mut new_stack = array![];
+        let mut pop_term: Pattern = bot();
+        let mut i = 0;
+
+        loop {
+            let term = self.elements.pop_front();
```
yzhang90, line 113: `pop_front` or `pop_front_consume`？

Robertorosmaninho: Pop_front.

cairo/src/stack.cairo
Robertorosmaninho: I don't expect it ever to happen. This is just something that Cairo forces us to use. We'll never push a `None` as it this is structure isn't present on the original proof checker.

yzhang90 (approved):
LGTM

Issue #626, opened 2023-12-14
Implement Instructions to "prove" phi -> phi in New Cairo


Robertorosmaninho:
## Instructions needed for phi -> phi
- [x] Implication
- [x] Prop1
- [x] Prop2
- [x] Save
- [x] Load
- [x] Publish
- [x] CleanMetaVar
- [ ] Modus Ponens
- [ ] Instantiate (Implication, MetaVar)

PR #632, opened 2023-12-14
Cairo checker execute instructions 2

Robertorosmaninho:
Part of #626 
This PR implements the following `execute_instructions`:
- CleanMetaVar
- Implies
- Instantiate
- ModusPonens 
- Prop1
- Prop2
- Prop3
- Save
- Load
- Pop
- Publish

Pattern Auxiliary Functions:
- PatternTrait
-  PatternTraitImpl skeleton

Lib Auxiliary Functions:
- instantiate_in_place
- instantiate_internal
    - Implies
    - MetaVar

Unit Tests:
- test_construct_phi_implies_phi
- test_save_and_load
- test_instantiate
- test_impreflex_compress_goal

Using `#[derive(Debug,...)` on Pattern and Term


cairo/src/verifier.cairo
```diff
@@ -336,10 +343,55 @@ fn execute_instructions(
                     Instruction::KnasterTarski => { panic!("KnasterTarski not implemented!"); },
                     Instruction::Instantiate => { panic!("Instantiate not implemented!"); },
                     Instruction::Pop => { panic!("Pop not implemented!"); },
-                    Instruction::Save => { panic!("Save not implemented!"); },
-                    Instruction::Load => { panic!("Load not implemented!"); },
-                    Instruction::Publish => { panic!("Publish not implemented!"); },
-                    Instruction::CleanMetaVar => { panic!("NoOp not implemented!"); },
+                    Instruction::Save => {
+                        let term = stack.last();
```
yzhang90, line 347: can we use `expect` here?

cairo/src/verifier.cairo
```diff
@@ -336,10 +343,55 @@ fn execute_instructions(
...
+                            },
+                            Option::None => { panic!("Save needs term on the Stack"); },
+                        }
+                    },
+                    Instruction::Load => {
+                        let index: u32 = buffer
+                            .pop_front()
+                            .expect('Error on Load instruction')
+                            .into();
+                        match memory.get(index).expect('Load needs term on Memory').unbox() {
```
yzhang90, line 363: The error msg is different from the reference impl.

cairo/src/verifier.cairo
```diff
@@ -336,10 +343,55 @@ fn execute_instructions(
...
+                            Term::Pattern(p) => stack.push(Term::Pattern(p.clone())),
```
yzhang90, line 364: Since we don't differentiate the Entry and Term, why not just directly push the clone of the memory in to the stack?

cairo/src/verifier.cairo
Robertorosmaninho: Message in expect must be short due to a current limitation of Cairo. It must be the size of felt252, which seems to be a few characters

yzhang90: ok

yzhang90 (approved):
LGTM

PR #633, opened 2023-12-15
Implement `ExecutionProofExp.prove_lift_through_rewrite`

### LINKED ISSUE #513: Implement `ExecutionProofExp.prove_lift_through_rewrite`
1. Assume axiom `p1 k= p2 -> (next(phi0[p1/x])  k= next(phi0[p2/x]))`

2. Add this equality axiom `p1 k= p2 -> ((psi -> p1') -> (psi -> p2'))`

3. Define a function:

   ```
   p1 k= p2          phi0  k=> phi1[p1/x]
   -----------------------
   phi0  k=> phi1[p2/x]
   ```

   where $x$ is an `app_ctx_hole` in `phi0`.

malturki:
~~This PR depends on #630~~

The PR is now ready. This closes #513.


generation/src/proof_generation/proofs/kore.py
```diff
@@ -241,7 +243,7 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
 reduce_top_axiom = Implies(kore_implies(phi0, kore_top(phi0), phi1), phi1)
 
 # TODO: Prove the axiom
-# (phi2:{phi0} k= phi3:{phi0}):{phi1} -> ( ♦(phi4[phi2/x]):{phi0} k= ♦(phi4[phi3/x]):{phi0} ):{phi1}
+# (phi2:{phi0} k= phi3:{phi0}):{phi1} k-> ( ♦(phi4[phi2/x]):{phi0} k= ♦(phi4[phi3/x]):{phi0} ):{phi1}
 keq_next_substitution_axiom = Implies(
```
nishantjr, line 247: Did you mean to change this to `kore_implies`?

generation/src/proof_generation/proofs/kore.py
```diff
@@ -252,6 +254,17 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
     ),
 )
 
+# TODO: Prove the axiom
+# (phi2:{phi0} k= phi3:{phi0}):{phi1} k-> ((phi5 k-> phi2):{phi4} k-> (phi5 k-> phi3):{phi4}):{phi6}
+keq_implication_axiom = Implies(
```
nishantjr, line 259: Same. Comment vs definition are different.

generation/src/proof_generation/proofs/kore.py
fiedlr: No, this should be implies because MP will be applied.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -227,6 +229,29 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
...
+        kore_next(MetaVar(4, app_ctx_holes=(EVar(0),)).apply_esubst(0, phi2)),
+        kore_next(MetaVar(4, app_ctx_holes=(EVar(0),)).apply_esubst(0, phi3)),
+    ),
+)
+
+# TODO: Prove the axiom
+# (phi2:{phi0} k= phi3:{phi0}):{phi1} k-> ((phi5 k-> phi2):{phi4} k-> (phi5 k-> phi3):{phi4}):{phi6}
+keq_implication_axiom = Implies(
+    kore_equals(phi0, phi1, phi2, phi3),
+    kore_implies(
```
fiedlr, line 248:
```suggestion
    Implies(
```

I think this should be `Implies`

generation/src/proof_generation/proofs/kore.py
```diff
@@ -227,6 +229,29 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
...
+    kore_equals(
+        phi0,
+        phi1,
+        kore_next(MetaVar(4, app_ctx_holes=(EVar(0),)).apply_esubst(0, phi2)),
+        kore_next(MetaVar(4, app_ctx_holes=(EVar(0),)).apply_esubst(0, phi3)),
+    ),
+)
+
+# TODO: Prove the axiom
+# (phi2:{phi0} k= phi3:{phi0}):{phi1} k-> ((phi5 k-> phi2):{phi4} k-> (phi5 k-> phi3):{phi4}):{phi6}
```
fiedlr, line 245:
```suggestion
# (phi2:{phi0} k= phi3:{phi0}):{phi1} k-> ((phi5 k-> phi2):{phi4} -> (phi5 k-> phi3):{phi4})
```

generation/src/proof_generation/proofs/kore.py
```diff
@@ -355,6 +382,57 @@ def reduce_top_in_imp(self, phi: ProofThunk):
...
+        # conclude: next(phi1[p1/x])  k= next(phi1[p2/x])
+        eq_next = self.modus_ponens(
+            self.dynamic_inst(
+                self.load_axiom(keq_next_substitution_axiom), {0: inner_sort, 1: outer_sort, 2: p1, 3: p2, 4: phi1}
+            ),
+            equality,
+        )
+
+        # MP on "Axiom: p1 k= p2 -> ((psi k-> p1) -> (psi k-> p2))" and "eq_next: next(phi1[p1/x])  k= next(phi1[p2/x])", with subst p1 |-> next(phi1[p1/x]), p2 |-> next(phi1[p2/x]), psi |-> phi0
+        # conclude: ((phi0 -> (next(phi1[p1/x])) k-> (phi0 -> next(phi1[p2/x]))))
```
fiedlr, line 422:
```suggestion
        # conclude: ((phi0 k-> (next(phi1[p1/x])) -> (phi0 k-> next(phi1[p2/x]))))
```

generation/src/proof_generation/proofs/kore.py
```diff
@@ -355,6 +382,57 @@ def reduce_top_in_imp(self, phi: ProofThunk):
...
+        inner_sort, outer_sort, p1, p2 = kore_equals.assert_matches(eq_next.conc)
+        phi0_imp_imp = self.modus_ponens(
+            self.dynamic_inst(
+                self.load_axiom(keq_implication_axiom),
+                {0: inner_sort, 1: outer_sort, 2: p1, 3: p2, 4: phi0, 5: outer_sort},  # Not sure what sort for 5!!
```
fiedlr, line 427:
```suggestion
                {0: inner_sort, 1: outer_sort, 2: p1, 3: p2, 4: phi0, 5: outer_sort},  # Not sure what sort for 5!!
```

See my comment above, no sort needed, as this needs to be a normal implication for MP

generation/src/proof_generation/proofs/kore.py
```diff
@@ -355,6 +382,57 @@ def reduce_top_in_imp(self, phi: ProofThunk):
...
+        #     ==(d)=>> next(phi1[p1/x])  k= next(phi1[p2/x])
+
+        # - MP on 2 and d, with subst p1 |-> next(phi1[p1/x]), p2 |-> next(phi1[p2/x]), psi |-> phi0
+        #     ==(e)=>> ((phi0 -> (next(phi1[p1/x])) k-> (phi0 -> next(phi1[p2/x]))))
+
+        # - MP on e and b, with subst identity
+        #     ==(f)=>> phi0 -> next(phi1[p2/x]))
+
+        inner_sort, outer_sort, p1, p2 = kore_equals.assert_matches(equality.conc)
+        imp_sort, imp_left, imp_right = kore_implies.assert_matches(imp.conc) # Not needed?
```
fiedlr, line 410:
```suggestion
        _, _, _ = kore_implies.assert_matches(imp.conc) # Not needed?
```
Or use the corresponding method that just asserts but does not destruct

generation/src/proof_generation/proofs/kore.py
```diff
@@ -227,6 +229,29 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
...
+        kore_next(MetaVar(4, app_ctx_holes=(EVar(0),)).apply_esubst(0, phi3)),
+    ),
+)
+
+# TODO: Prove the axiom
+# (phi2:{phi0} k= phi3:{phi0}):{phi1} k-> ((phi5 k-> phi2):{phi4} k-> (phi5 k-> phi3):{phi4}):{phi6}
+keq_implication_axiom = Implies(
+    kore_equals(phi0, phi1, phi2, phi3),
+    kore_implies(
+        phi6,
```
fiedlr, line 249:
```suggestion
```

fiedlr (changes requested):
Looks good, Musab! Please fix the remaining issues I highlighted and add one unit test with a single instance. It didn't work probably because MP is being applied on `k->`. The rest is clear, sorts are kept throughout the implications.

generation/src/proof_generation/proofs/kore.py
malturki: I suppose you meant that the topmost implication is not a kore_implies, right? So the comment should be:
```
(phi2:{phi0} k= phi3:{phi0}):{phi1} -> ((phi5 k-> phi2):{phi4} k-> (phi5 k-> phi3):{phi4}):{phi6}
                                   ^^^^
```
(which also means we would need `phi6` because kore_implies is sorted)

generation/src/proof_generation/proofs/kore.py
malturki: Yes, the comment was incorrect. It looks like we will need to review the comments in this file (in a future PR) as there are currently similar issues in other definitions.

generation/src/proof_generation/proofs/kore.py
malturki: Oh, nevermind. I see in your later comment that you expect both the outermost and the innermost implications not to be notations.

generation/src/proof_generation/proofs/kore.py
malturki: This is actually not needed, but I kept it there while I was building the proof. It's been removed.

malturki: Never mind. This became needed again after removing the extraneous parameters in this function. This is fixed in the latest commits.

fiedlr (approved):
Good job, Musab!

Two points:
a) We'll need to make a version that supports any hole `x`. But it's ok to suppose that it has id `0` for now, this is also done by `SimplificationPerformer`.
b) We can significantly simplify the proof with the refactorings proposed in https://github.com/orgs/runtimeverification/projects/47/views/1?pane=issue&itemId=47539153 
Even now, we can use `equality_with_subst` instead of adding the `keq_next...` axiom, but let's leave it like this for now since we will replace `equality_with_subst` with an even more general thing later.
c) I do not think we can reuse existing variables instead of having the matching on line 450, but let's leave it for now

PR #638, opened 2023-12-15
Sync wellformedness checking with R0 implementation and add wellformedness checking for Notation

fiedlr:
1. Updates the state of well_formedness checking in the DSL with that in the R0 proof checker using `__post_init__`
2. Adds `occurring_vars()` method on Patterns that keeps track of free variables actually occurring in the pattern, where metavars have no free variables unless instantiated with patterns containing them
3. Adds well_formedness checking to `Notation` that asserts the definition to contain no free variables

*Code review*
- Please think twice why I added the checkings on redundant ESubst/SSubsts before arguing this complicates things in the DSL. It is actually making things simple, *exactly* because we can make this assumption to remove edge cases like the ones I removed from the test suite. This is not only the case of the `metavars` method, but also the correctness of `e_fresh` and others depends on the substitutions not being redundant. 

     We should add a follow-up task that replaces `ESubst/SSubst` constructors with `apply_esubst/apply_ssubst` everywhere where it is applicable.

- I know that it is impossible to compute free variables guarantees correctly because of metavars, that is why we have freshness constraints. I'm open to rename `occurring_vars` to whatever as long as it is doing what I intended: keeping track of actual free occurences *currently* in the pattern.  

   I added a  note that `fv` makes no guarantees about freshness wrt proving soundness of the proof system, that's not its purpose.


generation/src/proof_generation/pattern.py
```diff
@@ -84,6 +84,14 @@ def evar_is_fresh(self, name: int) -> bool:
     def metavars(self) -> set[int]:
         raise NotImplementedError
 
+    """
+        Returns the set of all free variables occurring in the pattern
+        Makes no guarantees about freshness!
+    """
+
+    def fv(self) -> set[EVar | SVar]:
```
nishantjr, line 92: Lets rename this to `occuring_vars`. Otherwise it will be mistaken for `free_var` which has a different mathematical definition. It may also easily be mistaken for `fresh_var` so prefer full forms.

```suggestion
    """
        Returns the set of concrete variables occurring free in the pattern.
        Note that this is not the same as the set of free variables as it does
        reason about metavariables.
    """

    def occuring_vars(self) -> set[EVar | SVar]:
```

generation/src/proof_generation/pattern.py
```diff
@@ -415,13 +456,23 @@ def __str__(self) -> str:
...
+    Class to represent pending evar substitutions
+    It should almost never be used directly, instead use apply_esubst!
+"""
+
 
 @dataclass(frozen=True)
 class ESubst(Pattern):
     pattern: MetaVar | ESubst | SSubst
     var: EVar
     plug: Pattern
```
nishantjr, line 469:
```suggestion

@dataclass(frozen=True)
class ESubst(Pattern):
    """
    Represents evar substitutions over meta-variables.
    It should almost never be used directly, instead use apply_esubst!
    """

    pattern: MetaVar | ESubst | SSubst
    var: EVar
    plug: Pattern
```

Docstrings go inside the thing they document.

generation/src/proof_generation/pattern.py
```diff
@@ -415,13 +456,23 @@ def __str__(self) -> str:
...
 
+    def __post_init__(self) -> None:
+        # Check that ESubst is not redundant
+        assert self.var != self.plug
+        assert not self.pattern.evar_is_fresh(self.var.name)
```
nishantjr, line 474: This is not a check that the checker performs. Indeed it is expensive to do. If we keep this here we lose the ability to represent those proofs.

```suggestion
```

generation/src/proof_generation/pattern.py
```diff
@@ -450,19 +504,34 @@ def __str__(self) -> str:
...
+    Class to represent pending svar substitutions
+    It should almost never be used directly, instead use apply_ssubst!
+"""
+
+
 @dataclass(frozen=True)
 class SSubst(Pattern):
     pattern: MetaVar | ESubst | SSubst
     var: SVar
     plug: Pattern
```
nishantjr, line 517:
```suggestion


@dataclass(frozen=True)
class SSubst(Pattern):
    """
    Represent svar substitutions over meta-variables.
    It should almost never be used directly, instead use apply_ssubst!
    """

    pattern: MetaVar | ESubst | SSubst
    var: SVar
    plug: Pattern
```

nishantjr (approved):
Looks good, after the suggestions.

generation/src/proof_generation/pattern.py
nishantjr: Oh actually it isn't too bad, because its only over metavars.

fiedlr: It is implemented in the checker. 

The only issue here is that it actually does not remove all the wrong edge cases. We would need something like a dual to `e_fresh` arrays, that would enforce a variable to be free in any instantiation of a metavar and then do this (by `evar_is_free` I do not mean how `fv` is now implemented):

`assert self.pattern.evar_is_free(self.var.name)`

generation/src/proof_generation/pattern.py
fiedlr: Sounds good.

generation/src/proof_generation/pattern.py
fiedlr: Ooops, my bad

generation/src/proof_generation/pattern.py
fiedlr: Let's worry about this later

PR #642, opened 2023-12-18
Organize proof gen

nishantjr:
This PR cleans up the top-level source directory by organizing source files into folders.

It also includes shortcuts for importing from the `proof_generation.aml` and the `proof_generation.interpreter` modules. 

Theres some additional organization that can happen, but I decided to timebox it.


fiedlr (approved):
Looks good, Nishant, except for the mutual dependency `notation.py <-> syntax.py`. 

Not sure if this is ok with Python module system. If it is, then let's merge

nishantjr:
@fiedlr MyPy seems to handle that correctly. I think its doing two passes over the imports and treats typing imports differently.

Issue #653, opened 2023-12-19
Implement `instantiate_internal` for cairo proof checker


yzhang90:
- [x] apply_esubst
- [x] apply_ssubst
- [x] instantiate_internal::Evar
- [x] instantiate_internal::Svar
- [x] instantiate_internal::Symbol
- [x] instantiate_internal::Implies
- [x] instantiate_internal::App
- [x] instantiate_internal::Exists
- [x] instantiate_internal::Mu
- [x] instantiate_internal::MetaVar
- [x] instantiate_internal::ESbust
- [x] instantiate_internal::SSbust

PR #654, opened 2023-12-19
Implement apply_esubst and apply_ssubst

yzhang90:
Part of #653

### DIFF
--- a/cairo/src/verifier.cairo
+++ b/cairo/src/verifier.cairo
@@ -273,6 +273,54 @@ fn apply_esubst(pattern: @Pattern, evar_id: Id, plug: @Pattern) -> Pattern {
     }
 }
 
+fn apply_ssubst(pattern: @Pattern, svar_id: Id, plug: @Pattern) -> Pattern {
+    let wrap_subst = ssubst(pattern.clone(), svar_id, plug.clone());
+
+    match pattern {
+        Pattern::SVar(e) => { if *e == svar_id {
+            plug.clone()
+        } else {
+            pattern.clone()
+        } },
+        Pattern::Implies(ImpliesType{left,
+        right }) => {
+            let left = left.clone().unwrap().unbox();
+            let right = right.clone().unwrap().unbox();
+
+            implies(apply_ssubst(@left, svar_id, plug), apply_ssubst(@right, svar_id, plug))
+        },
+        Pattern::App(AppType{left,
+        right }) => {
+            let left = left.clone().unwrap().unbox();
+            let right = right.clone().unwrap().unbox();
+
+            app(apply_ssubst(@left, svar_id, plug), apply_ssubst(@right, svar_id, plug))
+        },
+        Pattern::Exists(ExistsType{var,
+        subpattern }) => {
+            let subpattern = subpattern.clone().unwrap().unbox();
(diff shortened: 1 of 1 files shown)


cairo/src/verifier.cairo
```diff
@@ -220,6 +222,105 @@ fn from(value: felt252) -> Instruction {
...
+        Pattern::SSubst(_) => wrap_subst,
+        _ => pattern.clone(),
+    }
+}
+
+fn apply_ssubst(pattern: @Pattern, svar_id: Id, plug: @Pattern) -> Pattern {
+    let wrap_subst = ssubst(pattern.clone(), svar_id, plug.clone());
+
+    match pattern {
+        Pattern::SVar(e) => { if *e == svar_id {
```
Robertorosmaninho, line 280: Not a big deal, but could we make it `Pattern::SVar(s)`?

cairo/src/verifier.cairo
```diff
@@ -220,6 +222,105 @@ fn from(value: felt252) -> Instruction {
...
+            plug.clone()
+        } else {
+            pattern.clone()
+        } },
+        Pattern::Implies(ImpliesType{left,
+        right }) => {
+            let left = left.clone().unwrap().unbox();
+            let right = right.clone().unwrap().unbox();
```
Robertorosmaninho, line 288: We should keep a pattern here. Before, you used `left_unbox` for these situations, and now you're using `left`. Honestly, I prefer this pattern of overwriting the variable name just to make things cleaner, but we should pick one and stick to it.

cairo/src/verifier.cairo
```diff
@@ -220,6 +222,105 @@ fn from(value: felt252) -> Instruction {
...
+    match pattern {
+        Pattern::EVar(e) => { if *e == evar_id {
+            plug.clone()
+        } else {
+            pattern.clone()
+        } },
+        Pattern::Implies(ImpliesType{left,
+        right }) => {
+            let left = left.clone().unwrap().unbox();
+            let right = right.clone().unwrap().unbox();
```
Robertorosmaninho, line 240: Same about `left` and `left_unbox`...

cairo/src/verifier.cairo
```diff
@@ -220,6 +222,105 @@ fn from(value: felt252) -> Instruction {
     }
 }
 
+/// Substitution utilities
+/// ----------------------
+
+fn apply_esubst(pattern: @Pattern, evar_id: Id, plug: @Pattern) -> Pattern {
+    let wrap_subst = esubst(pattern.clone(), evar_id, plug.clone());
```
Robertorosmaninho, line 229: What is the difference between this and ` let wrap_subst = || esubst(Rc::clone(pattern), evar_id, Rc::clone(plug));` regarding the `= ||`?

Robertorosmaninho (approved):
Approving with some comments.

cairo/src/verifier.cairo
yzhang90: `let wrap_subst = || esubst(Rc::clone(pattern), evar_id, Rc::clone(plug));` this is a constant function, it takes a zero argument and return the esubst pattern. I think they use this to avoid clone?

PR #656, opened 2023-12-19
`ProofExp.pretty_diff` Tool for diffing patterns.

nishantjr:
* This is `Instantiation` aware
* It reuses the pretty printing code through the use a `Diff` subclass of `Pattern`. This is a little too "clever" for my liking but greatly simplifies the code.
* I like the output for the most part. One area that is lacking is indentation. We'd need to change pretty printing itself to be notation aware to support that.
* This can be called without a `ProofExp`, but would require manually specifying `PrettyOptions`.


generation/src/tests/unit/test_utils.py
```diff
@@ -68,3 +70,88 @@ def test_match() -> None:
...
+        '--- phi0\n'
+        '+++ phi1\n'
+        ' -> \n'
+        '--- phi1\n'
+        '+++ phi2\n'
+        ')')
+
+    # Exists
+    assert exp.pretty_diff(Exists(0, phi1), Exists(0, phi1)) == '(∃ x0 . phi1)'
+    assert exp.pretty_diff(Exists(0, phi1), Exists(1, phi1)) == '\n--- (∃ x0 . phi1)\n+++ (∃ x1 . phi1)\n'
```
fiedlr, line 105: Maybe here we should just return `exists -- x0 ++ x1 . phi1`?

fiedlr (approved):
Looks great, Nishant! Maybe we could add some location info in some other prs. I left one small comment.

PR #657, opened 2023-12-19
Complete the `instantiate_internal` function

yzhang90:
Complete #653

### DIFF
--- a/cairo/src/verifier.cairo
+++ b/cairo/src/verifier.cairo
@@ -348,9 +348,50 @@ fn instantiate_internal(
                 return Option::Some(implies(inst_left.unwrap(), inst_right.unwrap()));
             }
         },
-        Pattern::App(_) => Option::None,
-        Pattern::Exists(_) => Option::None,
-        Pattern::Mu(_) => Option::None,
+        Pattern::App(AppType{left,
+        right }) => {
+            let mut left = left.unwrap().unbox();
+            let mut right = right.unwrap().unbox();
+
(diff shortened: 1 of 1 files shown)


cairo/src/verifier.cairo
```diff
@@ -348,21 +348,61 @@ fn instantiate_internal(
...
+            }
+        },
+        Pattern::Exists(ExistsType{var,
+        subpattern }) => {
+            let mut subpattern = subpattern.unwrap().unbox();
+
+            let mut new_sub = instantiate_internal(ref subpattern, ref vars, ref plugs);
+
+            if new_sub.is_none() {
+                Option::None
```
Robertorosmaninho, line 378: Cairo must return None here?

cairo/src/verifier.cairo
```diff
@@ -348,21 +348,61 @@ fn instantiate_internal(
...
+            }
+        },
+        Pattern::Mu(MuType{var,
+        subpattern }) => {
+            let mut subpattern = subpattern.unwrap().unbox();
+
+            let mut new_sub = instantiate_internal(ref subpattern, ref vars, ref plugs);
+
+            if new_sub.is_none() {
+                Option::None
```
Robertorosmaninho, line 390: Same question here.

Robertorosmaninho (approved):
Approved with some questions

cairo/src/verifier.cairo
yzhang90: The reference impl is
```
Some(exists(*var, new_sub?))
```
That  ? means if new_sub is none, then the whole expression is None.

Issue #624, opened 2023-12-13
Update zkLLVM to its latest version


Robertorosmaninho:
The ;=nil Foundation team has fixed the issue that we reported in zkLLVM: https://github.com/NilFoundation/zkLLVM/issues/412

They now only support LLVM 17, so we should:
- [ ] Update our scripts to work with latest version of zkLLVM
- [ ] Check if we need to update our readme 

The latest version of `assigner` can't compile our proof checker, but can compile the CSL example, so we should try to fix it on our end.
- [ ] Update proof-checker to work with laster version of zkLLVM
- [ ] Now that we have the latest version working we can re-generate the numbers for the CSL Examples in zkLLVM
- [ ] Now that we have the lastest version of zkLLVM we can set out CI:

PR #660, opened 2023-12-20
Using `std::memset` instead of our own `memset_` implementation.

Robertorosmaninho:
This PR is part of #624!
It updates our zkllvm proof-checker to work with the new version of zkllvm. It's now possible to run assigner again without breaking with error (besides an eventual lack of memory);


yzhang90 (approved):
LGTM

PR #663, opened 2023-12-20
Disable stack printing in pretty proof files

MirceaS:


generation/src/proof_generation/k/proof_gen.py
```diff
@@ -41,7 +41,7 @@ def generate_proof_file(proof_gen: ProofExp, output_dir: Path, slice_name: str,
     """Generate the proof files."""
     if not output_dir.exists():
         output_dir.mkdir(parents=True)
-    mode = 'pretty' if pretty else 'binary'
+    mode = 'pretty-no-stack' if pretty else 'binary'
```
fiedlr, line 44: Please change the var name here as we discussed

fiedlr (approved):

PR #658, opened 2023-12-19
Add support for lifting simplification proofs to rewrites

fiedlr:
- Parametrizes `node_tree` semantics with the output sort
- Changes `SimplificationPerformer` to produce equalities with the topcell output sort
- Adds lifting to `simplification_event`, we now have a functional Rewrite+Simplification Batch pipeline at least for one  rewrite
- Adds a `HOLE` constant to denote `EVar`'s meant to be used as an app ctx hole
- Adds `pretty_diff` on patterns to several asserts to simplify debugging
- Adds outer_sort parametrization for `sorted_eq_id`
- `subst_in_rewrite_target` now requires the resulting configuration to be specified by `phi1` 
- Fixes parametrization of the `subpattern_batch` test where `DummyProver` was also plugged in to the main performer

PR #664, opened 2023-12-20
Add code for processing functional events and tests for traces with them

### LINKED ISSUE #646: Add simplification events to `rewrite_steps.py` and `execution_proof_generation.py`
# Intro
Simplification events are parsed in the hint files, but currently not processed further

Depends on #645 

# Tasks
- [ ] Update `rewrite_steps.py` to recognize simplification events and yield pre-processed hints for them called `SimplificationStepExpression`. The processed hints should include all the arguments of `simplification_event` with the same types
- [ ] Add tests that a) pre-processed hints are correct 
- [ ] Add unit tests with hardwired hints similar to `test_double_rewrite_semantics`

IljaZakharov:
The PR adds:

- Processing of functional events;
- Tests for hint iterator for the tree-reverse example without integers;
- Tests for llvm trace parser for the tree-reverse example without integers.

Closes #646


generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
```diff
@@ -69,19 +61,30 @@ def get_proof_hints(
...
+    objects_yielded = 0
     for event in rewrite_trace.trace:
         # TODO: process function/hook events in llvm_proof_hint.strace
+        hints_processed += 1
         match event:
             case LLVMRuleEvent(rule_ordinal, substitutions):
                 axiom = language_semantics.get_axiom(rule_ordinal)
                 converted_substitutions = language_semantics.convert_substitutions(dict(substitutions), rule_ordinal)
-                hint = RewriteStepExpression(axiom, converted_substitutions)
+                hint = RewriteStepExpression(axiom, frozendict(converted_substitutions))
```
fiedlr, line 79: This variable is not needed? We can pass the term to yield

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
```diff
@@ -69,19 +61,30 @@ def get_proof_hints(
...
                 yield hint
+            case LLVMFunctionEvent(name, location_str, _):
+                location: tuple[int, ...] = tuple(map(int, location_str.split(':')))
+                objects_yielded += 1
+                yield FunEvent(name, location)
             case _:
                 continue
+    else:
+        if objects_yielded == 0:
+            print(f'WARNING: No hints generated for a trace with {hints_processed} events')
```
fiedlr, line 90: This should be an `assert`?

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
```diff
@@ -69,19 +61,30 @@ def get_proof_hints(
...
     rewrite_trace: LLVMRewriteTrace,
     language_semantics: LanguageSemantics,
-) -> Iterator[RewriteStepExpression]:
+) -> EventTrace:
     """
     Emits proof hints corresponding to the given LLVM rewrite trace.
     Note that no hints will be generated if the trace is empty.
     """
     # TODO: process function/hook/rule events in llvm_proof_hint.pre_trace
+    hints_processed = 0
```
fiedlr, line 70: This one is not used anywhere?

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
```diff
@@ -69,19 +61,30 @@ def get_proof_hints(
...
+    objects_yielded = 0
```
fiedlr, line 71: This can be just a boolean `nonempty_yield`?

generation/src/tests/integration/rewrite/test_llvm_proof_generator.py
```diff
@@ -42,18 +41,22 @@ def get_k_cell_top_symbol(term: Pattern) -> Pattern:
...
-    if isinstance(term, Instantiate):
-        term = term.simplify()
-    return term
+    if isinstance(term, App):
+        term = term.right
+        if isinstance(term, Instantiate):
+            term = term.simplify()
+        return term
+    elif isinstance(term, Symbol):
+        return term
```
fiedlr, line 50: Was this case forgotten? Is this function covering all cases?

fiedlr (approved):
Looks great, Ilja! Thanks. Just left a couple of comments.

-- Please merge after https://github.com/runtimeverification/proof-checker/pull/658 -- , as it is longer and it might be difficult to add with this change present

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
IljaZakharov: Fixed!

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
IljaZakharov: I can do that, sure. It depends on actually expected behaviour for some tests. Currently, I've removed that `simple.k` file because it cannot pass with its trace containing unsupported events. But we used the warning before, as it had an empty trace because we ignored functional simplification and hooks and everything else.

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
IljaZakharov: It is used in the assertion warning.

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
IljaZakharov: Fixed.

generation/src/tests/integration/rewrite/test_llvm_proof_generator.py
IljaZakharov: This is just a helper function for reaching the K cell content. But this path is a bit different in rewrite examples and in the tree-reverse example. It is hardly possible to make a function that would find this way for any possible example.

Issue #628, opened 2023-12-14
Summary of Potentially Relevant K Features Used by KEVM


malturki:
The following lists K features that could potentially be relevant for proof generation based on concrete execution with the LLVM backend. Some of these features are not semantical but are included anyway for completeness.

Note: this is based on [commit f74bb9 of evm-semantics](https://github.com/runtimeverification/evm-semantics/tree/f74bb95a736cba96fd532b7ccc59f48c9b6cf57b).

## Modules

The following module importations and annotations are used:

- Imports several built-in modules, including `BOOL`, `INT`, `BYTES`, `STRING`, `LIST`, `SET`, `MAP`, `K-EQUAL`, `K-IO` (in `JSON-RPC`), ... etc.
- Imports `KRYPTO` (in `SERIALIZATION`), which defines several (hooked) cryptographic hash functions, such as `keccak256`
- Some modules are decorated with:
    - the`kore` attribute (in the `optimizations.md` file)
    - the `symbolic` attribute (in the `lemmas` subdirectory, e.g. the `GAS-SIMPLIFICATION` module). As per the K manual, these modules are ignored in concrete executions with the LLVM backend.

## Built-in Data Types

The following built-in data types are used:

- Several built-in types and their (hooked) operations, including: `Bool` (inc. built-in equality, built-in conditional, ...), `Int`, `Bytes`, `String` (inc. substrString, +String, ...), `KResult`, collection types, ... etc.
- All three types of lists:
    1. User-defined lists:
        - Associative lists (e.g. `syntax OpCodes ::= ".OpCodes" | OpCode ";" OpCodes`)
        - Cons lists (e.g. `syntax WordStack ::= ".WordStack" | Int ":" WordStack`)
    2. Lists constructed using `List(Type, sep)`
    3. Lists constructed using the generic List type (`ListItem`, `.List`, ...)
- Sets and Maps are constructed using the generic `Set` and `Map` types
- An extended implementation of JSON

## Syntax Productions

The following syntax attributes/annotations are used:
- `klabel` (although no two constructors share the same label as far as I could tell)
- `symbol`
- `function`
- `total`
- `macro`
- `alias`
- `no-evaluators`
- `smtlib(...)`
- `strict`, `seqstrict`
- `left`
- `memo` (e.g. `#computeValidJumpDests(Bytes)`)
- `priority` (used only in the `optimizations.md` file and the lemmas subdirectory)
- precedence indicator `>`

## Rules

[...]

PR #668, opened 2023-12-21
Add new K tests based on K features used in KEVM

### LINKED ISSUE #651: Augment the tests list with minimal representative examples based on K features used in KEVM

malturki:
This PR adds a bunch of K tests inspired by the features of KEVM extracted and documented [here](https://github.com/runtimeverification/proof-checker/issues/628). It closes #651.

The PR makes primarily the following changes:
1. Adds new tests (some of which are imported from [here](https://github.com/runtimeverification/proof-generation/tree/main/examples), while the others are entirely fresh)
2. Improves formatting and adds rule labels to most existing tests
3. Disables tests in `Makefile` that are currently failing
4. Updates snapshots and python tests

The following is a list of tests added or modified:

1.  Basic Rewriting

    1.  [`single-rewrite.k`]: One ground rewrite rule. No use of transitivity `=>*`
    2.  [`double-rewrite.k`]: Two ground rewrite rules, with transitivity of `=>*`
    3.  [`projection.k`]: One rewrite rule with variables. No use of transitivity `=>*`
    4.  [`decrement.k`]: One rewrite rule with variables, with and without transitivity
    5.  [`add-rewrite.k`]: Multiple rewrite rules with constants, variables, sequencing operator, and with transitivity

2.  Functional Simplification

    1. [`non-rec-function.k`]: Non-recursive (total) function evaluation
    2. [`custom-klabel-function.k`]: Non-recursive functions using various combinations of `klabel`/`symbol` attributes
    3. [`assoc-function.k`]: Left- and right-associative functions
    4. [`tree-reverse.k`]: Recursive (total) function evaluation
    5. [`memo-function.k`]: Memoized function evaluation

3.  Hooks/Domain Values

    1. [`builtin-int.k`] : Integer addition hook (no user-defined functions)
    2. [`builtin-functions.k`] : A collection of functions defined using hooked symbols from `BOOL`, `K-EQUAL`, `INT`, `BYTES`, and `STRING`
    3. [`builtin-io-fun.k`] : A module that reads from file using K-IO
    4. [`builtin-json-fun.k`] : A function that converts JSON to String and back using JSON

4.  Conditional Rules:

    1. [`two-counters.k`]: A conditional rewrite rule with hooks
    2. [`conditional-function.k`]: Conditional, recursive functional rewrite rules with hooks
    3. [`peano.k`]: Conditional rewrite rules using **strictness** attributes

5.  [`injections.k`]: Subsorting with matching modulo axioms such as `inj o inj = inj`

6.  Lists

    1. Manually defined lists
        * `list-assoc.k`]: Functions on an associative list (using the `right`attribute)
        * `list-cons.k`]: Fucntions on a cons list
    2. [`list-factory.k`]: Fucntions on a list constructed using the `List(Type, sep)` constructor
    3. [`list-semantic.k`]: Fucntions on a semantic list (constructed using the generic List type)

7.  Sets

    1. [`set-fun.k`]: Functions on a set (constructed using the generic Set type)

8.  Maps

    1.  [`map-fun.k`]: Functions on a map (constructed using the generic Map type, requiring support for isKResult, strictness and evaluation contexts)

9.  Cells and Configurations

    1. [`nested-cells.k`]: Declares a minimal definition with nested cells
    2. [`exit-cell.k`]: Declares a minimal definition with an `exit-code` cell
    3. [`modular-config.k`]: Declares a modular configuration specs with two sub-configurations
    4. [`cell-collection.k`]: Declares a collection cell with multiplicity `*` and type `Map`
    5. [`cell-value.k`]: Defines operations on cells as first-class values, including dynamic creation and deletion

10. Larger Tests with Different Combinations of Features

    1.  [`imp.k`]: The full `IMP` language
    2.  [`imp5.k`]: `IMP` with a fixed 5-slot state (no maps)
    3.  [`imp5-rw-succ`]: A stripped down version of `IMP5` that doesn't use K's builtin domain values
    4.  [`imp5-rw-literal`]: A further stripped down version of `IMP5` that doesn't use K's builtin domain values or strictness attributes
    5.  [`arith.k`]: Operations on builtin integers (also uses `context`)
    6.  [`lambda-explicit-substitution.k`]: Explicit lambda substitution functions (using `seqstrict`, `left`, `bracket`, domain values, side conditions, ...)
    7.  [`reg.k`]: An assembly language with various language features
    8.  [`pcf.k`]: A functional language with various language features

11. Other Attributes and Features (some are less important than others)
    1. [`prioritized-rule.k`]: Two overlapping rewrite rules with different priorities
    2. [`is-zero.k`]: An "otherwise" rule used in defining a function
    3. [`simplification.k`]: A function with `no-evaluators` used in two `simplification` rules
    4. [`macro.k`]: A function that refers to `macro` and `alias` symbols
    5. [`fun-context.k`]: Function rules referring to global context `[[ ... ]]`
    6. [`let.k`]: A function defined using the `#let` binding construct
    7. [`fresh-gen.k`]: Uses the fresh generator `!` on integers
    8. [`type-cast.k`]: Uses the type casting operator `:>`

## Notes

1. I will update the [test-based milestones issue](https://github.com/runtimeverification/proof-checker/issues/521) with update-to-date details about the tests once this PR is merged.
2. This PR increases CI time since there are many more tests to generate proof hints for. If the increase is significant, we might want to restrict CI to running only a subset by default. We could introduce a Make target that runs all the tests only on demand.


proofs/proof-hints/tree-reverse-without-integers/simplify.tree-reverse-without-integers.hints.pretty
```diff
@@ -1,3 +1,19 @@
...
+Function: Lblreverse'LParUndsRParUnds'TREE-REVERSE-WITHOUT-INTEGERS-SYNTAX'Unds'Tree'Unds'Tree{} @ 0:0
+  Term: [kore]
+Rule 155 [label: TREE-REVERSE-WITHOUT-INTEGERS.base-case-b] [loc: 15,26] [substs: 1]
+  Var'Unds'Gen0 = [kore]
+Function: Lblreverse'LParUndsRParUnds'TREE-REVERSE-WITHOUT-INTEGERS-SYNTAX'Unds'Tree'Unds'Tree{} @ 0:1
+  Term: [kore]
+Rule 154 [label: TREE-REVERSE-WITHOUT-INTEGERS.base-case-a] [loc: 14,26] [substs: 1]
+  Var'Unds'Gen0 = [kore]
+=======
+>>>>>>> main
```
nishantjr, line 16: Whats up with this?

nishantjr:
Great work Musab. Lets not add the pretty hints. I don't think anyone will look at them, and they can be generated pretty quicly? It may be more useful to add instructions in the README documenting how to generate them.

generation/k-benchmarks/builtin-io/#input_file
```diff
@@ -0,0 +1 @@
+The quick brown fox jumped over the lazy dog
```
nishantjr, line 1: Rename this file to not start with a sharp

nishantjr (approved):
Great work Musab. If it would take less than half an hour, could you move them into folders with a numbered prefix, based on the issue #521 ?

e.g.

```
00-basic-rewriting/00-single-rewrite/...
00-basic-rewritint/01-double-rewrite/...
```

proofs/proof-hints/tree-reverse-without-integers/simplify.tree-reverse-without-integers.hints.pretty
malturki: Wow! Nice catch. Pretty-printed hints were removed.

malturki:
Thanks, @nishantjr, for the review. I addressed all the comments, including adding more documentation, except the numbering of tests, for which I'll add an issue to be addressed in a future PR (this is because numbering will likely require modifying how the Makefile discovers and processes tests, which will probably require more than 30 minutes to complete and test fully).

PR #675, opened 2024-01-08
Add MetaML soundness file

fiedlr:
Introduces a proof of soundness for matching logic extended with metavariables for proofs with the following properties:
1. Every metavariable has hole constraints disjoint from other constraints.
2. Metavariables have consistent constraints across the whole proof (not just patterns).


nishantjr (approved):
Looks good. Lets add the compiled PDF though. Rename to `soundness.tex` since we decided against `MetaML`?

fiedlr:
> Looks good. Lets add the compiled PDF though. Rename to `soundness.tex` since we decided against `MetaML`?

Sounds good! Fixed, thanks.

PR #681, opened 2024-01-09
Preliminary definitions for kore functionality

MirceaS:


generation/src/proof_generation/proofs/definedness.py
```diff
@@ -10,12 +10,15 @@
 floor = Notation('floor', 1, neg(ceil(neg(phi0))), '⌊ {0} ⌋')
 subset = Notation('subset', 2, floor(Implies(phi0, phi1)), '({0} ⊆ {1})')
 equals = Notation('equals', 2, floor(equiv(phi0, phi1)), '({0} = {1})')
-functional = Notation('functional', 1, Exists(0, equals(EVar(0), MetaVar(0, (EVar(0),)))), 'functional({0})')
+
+
```
fiedlr, line 14: Functional also needs to be `cached`

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+def kore_forall(var: int) -> Notation:
+    """kore_forall(inner_sort, outer_sort, pattern)"""
+    return Notation(
+        'kore-forall',
+        3,
+        _and(sorted_forall(var)(phi0, phi2), kore_top(phi1)),
+        f'( k∀ {var}:{0} . {2}):{1}',
+    )
+
+
```
fiedlr, line 133: @cache

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+def kore_functional(var: int) -> Notation:
+    """kore_functional(var)(sort_of_pattern, outer_sort, pattern)"""
+    return Notation(
+        'kore-functional',
+        3,
+        kore_exists(var)(phi0, phi1, kore_equals(EVar(var), MetaVar(2, e_fresh=(EVar(var),)))),
+        'kore_is_functional({2}:{0}):{1}',
+    )
+
+
```
fiedlr, line 143: @cache

generation/src/proof_generation/proofs/definedness.py
```diff
@@ -10,12 +10,15 @@
...
+def functional(var: int = 0) -> Notation:
```
nishantjr, line 15: Lets remove this default. Its confusing otherwise.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -48,6 +48,13 @@ def sorted_exists(var: int) -> Notation:
     return Notation('sorted-exists', 2, Exists(var, _and(in_sort(EVar(var), phi0), phi1)), f'( ∃ x{var}:{0} . {1} )')
 
 
+@cache
+def sorted_forall(var: int) -> Notation:
```
nishantjr, line 52: Lets rename this to `kore_forall` for consistency.

nishantjr (approved):
Looks good, once suggestions applied.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+    """kore_forall(inner_sort, outer_sort, pattern)"""
+    return Notation(
+        'kore-forall',
+        3,
+        _and(sorted_forall(var)(phi0, phi2), kore_top(phi1)),
+        f'( k∀ {var}:{0} . {2}):{1}',
+    )
+
+
+def kore_functional(var: int) -> Notation:
```
fiedlr, line 134: Can't we use `functional` here somehow?

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+    """kore_functional(var)(sort_of_pattern, outer_sort, pattern)"""
```
fiedlr, line 135:
```suggestion
    """kore_functional(var)(inner_sort, outer_sort, pattern)"""
```

generation/src/proof_generation/proofs/kore.py
MirceaS: `kore_forall` is defined below. `sorted_forall` and `kore_forall` are different, the same way that we already have `sorted_exists` and `kore_exists`. The `kore_` version is the `sorted` version limited to the appropriate outer sort.

generation/src/proof_generation/proofs/kore.py
fiedlr: `kore_equals` is missing the input and outer sort here

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+    return Notation(
+        'kore-functional',
+        3,
+        kore_exists(var)(phi0, phi1, kore_equals(EVar(var), MetaVar(2, e_fresh=(EVar(var),)))),
+        'kore_is_functional({2}:{0}):{1}',
+    )
+
+
+def kore_is_function(arity: list[Pattern]) -> Notation:
+    """kore_is_function(arity)(sort_of_symbol, outer_sort, symbol)"""
```
fiedlr, line 145:
```suggestion
    """kore_is_function(arity)(inner_sort, outer_sort, symbol)"""
```

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+    """kore_functional(var)(sort_of_pattern, outer_sort, pattern)"""
+    return Notation(
+        'kore-functional',
+        3,
+        kore_exists(var)(phi0, phi1, kore_equals(EVar(var), MetaVar(2, e_fresh=(EVar(var),)))),
+        'kore_is_functional({2}:{0}):{1}',
+    )
+
+
+def kore_is_function(arity: list[Pattern]) -> Notation:
```
fiedlr, line 144:
```suggestion
def kore_is_function(arg_sorts: list[Pattern]) -> Notation:
```

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+    """kore_is_function(arity)(sort_of_symbol, outer_sort, symbol)"""
+    notation_pattern: Pattern
+    notation_pattern = MetaVar(2, e_fresh=tuple([EVar(x) for x in range(len(arity) + 1)]))
+    for index in range(len(arity)):
+        notation_pattern = App(notation_pattern, EVar(index))
```
fiedlr, line 149: On very fast glance, I'm not sure what is is doing. Why App?

fiedlr (changes requested):

generation/src/proof_generation/proofs/kore.py
nishantjr: OK.

generation/src/proof_generation/proofs/kore.py
MirceaS: Not really. ML `functional` uses a simple existential whereas kore `functional` has to use sorted existentials and kore equality instead.

generation/src/proof_generation/proofs/kore.py
MirceaS: `inner_sort` is very confusing imo. There are notations that have 0 or multiple "inner" sorts depending on which inner part you're looking at. I want to make it explicit that this has to be the sort of the pattern provided at the end.
The `inner_sort` terminology only really works for a few very simple constructs, like kore equality.

generation/src/proof_generation/proofs/kore.py
MirceaS: Ditto.

generation/src/proof_generation/proofs/kore.py
MirceaS: Given a pattern `pat`, what this does is construct the pattern `forall x0. forall x1. ... forall xn. exists y. y = pat x0 x1 ... xn` where the x's are arguments to the (usually) symbol `pat`.
This is what being a function means.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+    notation_pattern = kore_functional(len(arity))(phi0, phi1, notation_pattern)
+    for index, sort in enumerate(arity):
+        notation_pattern = sorted_forall(index)(sort, notation_pattern)
+    return Notation('kore-function', 3, notation_pattern, 'is_kore_function({2}:{0}):{1}')
```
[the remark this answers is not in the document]
MirceaS: `kore_functional` also has `is` in the printed form. Do you want me to change that also?

generation/src/proof_generation/proofs/kore.py
nishantjr: Then lets do `return_sort`.

MirceaS: @nishantjr the "return" sort is the `outer_sort`, not inner. I think that one is named appropriately.

generation/src/proof_generation/proofs/kore.py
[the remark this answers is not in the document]
nishantjr: Yes please, the names and the pretty printing should be consistent at the very least.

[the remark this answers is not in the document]
MirceaS: What about keeping the `is` in all of them, since they're all predicates?

generation/src/proof_generation/proofs/kore.py
fiedlr: Let's be consistent, we use inner_sort with everywhere else. We can rename it in a different pr.

generation/src/proof_generation/proofs/kore.py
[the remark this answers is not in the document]
fiedlr: Let's stick with what we have and worry about naming in a separate pr

generation/src/proof_generation/proofs/kore.py
fiedlr: Aha, I see, not seeing a symbol was confusing

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,44 @@ def kore_exists(var: int) -> Notation:
...
+    )
+
+
+def kore_is_function(arity: list[Pattern]) -> Notation:
+    """kore_is_function(arity)(sort_of_symbol, outer_sort, symbol)"""
+    notation_pattern: Pattern
+    notation_pattern = MetaVar(2, e_fresh=tuple([EVar(x) for x in range(len(arity) + 1)]))
+    for index in range(len(arity)):
+        notation_pattern = App(notation_pattern, EVar(index))
+    notation_pattern = kore_functional(len(arity))(phi0, phi1, notation_pattern)
```
fiedlr:
```suggestion
    notation_pattern = kore_functional(len(arity) + 1)(phi0, phi1, notation_pattern)
```

since last `x` is `len(arity)`, `y` should start at `len(arity) + 1`

fiedlr (approved):
Looks good, Mircea, thanks. Just fix the bug with id indexing and do the renamings suggested by us. We can have a separate PR that fixes the naming to `is_` for predicates.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -108,11 +115,47 @@ def kore_exists(var: int) -> Notation:
...
+        'kore_is_functional({2}:{0}):{1}',
+    )
+
+
+@cache
+def kore_is_function(arg_sorts: list[Pattern]) -> Notation:
+    """kore_is_function(arity)(sort_of_symbol, outer_sort, symbol)"""
+    arg_len = len(arg_sorts)
+    notation_pattern: Pattern
+    notation_pattern = MetaVar(2, e_fresh=tuple([EVar(x) for x in range(arg_len + 1)]))
```
fiedlr, line 150:
```suggestion
    notation_pattern = MetaVar(2, e_fresh=tuple([EVar(x) for x in range(arg_len)]))
```

We want to have exactly `arg_len` of x's, which means this line as we start from 0 and go to `arg_len - 1`.

Then `y` will start at `arg_len` as you are using it in `kore_functional`

MirceaS: What do you mean by `y`? EVars don't have string names, they just have integer id's. Only symbols have string names.
We need to account for the `arg_len` evars standing for the arguments as well as an extra one that is used to claim the functionality of the entire application.

PR #690, opened 2024-01-10
Add annotation instructions to interpreters for debugging

MirceaS:


fiedlr (approved):
Not sure how this helps with debugging, but I don't see any huge issues with adding it for now.

PR #683, opened 2024-01-09
Processing hook events (Part I)

IljaZakharov:
Add support for other types of simplification events: Hook events. Simplification is performed using the axiom added on the fly for domain values calculation carried out by K. 

Partially implements #459.

PR #691, opened 2024-01-10
Processing hook events  (Part II)

### LINKED ISSUE #459: Add whitelist assumptions for every hook event
Follow up of #458 .

# Intro 

Hook events provide results for various domain reasoning like basic arithmetic over integers for which we do not have explicit definitions.
Computations of these are delegated to the backend and  other tools for performance reasons. 

# Problem

Results of hook events will not be proved. Instead, we need to add for every such event an equality axiom of the form
```
hook(arg1, ..., argn) = result
```

where hook is a name of the domain function applied to arguments arg1...argn, equality is the notation from the definedness module, and result is given by the proof hint. 

This task adds these to Gamma and add simplification handling in the simplest form (top-level).

# Sketch of the solution

- [x] Add a `dv.k` test file and check that we can generate hints for the example.
- [x] Define a test semantics for the `dv.k` for further unit testing.
- [x] Add a function collect_whitelist_assumptions that will be similar to collect_functional_axioms. For every hook event of a rewrite step, it will be called and the assumptions added to axioms. This will be memorized to prevent duplicity.
- [x] Update the simplification Prover and Performer classes to generate proofs using new axioms.
- [x] Add a unit test with a rewrite step containing hook events and expected output for collect_functional_axioms
- [x] Update hints processing and add all new necessary classes to forward hook events to the Proof Expression class.
- [x] Add tests for hardwired hints trace for the `dv` example, as it is done for tree-reverse and simple-rewrite examples. We need to test that hints are processed correctly by the llvm hints parse and `get_proof_hints`
- [x] Add an integration test with a single rewrite doing a single addition and prove it with the added axiom.

Example of a trace with hook events (for `dv.k` and `succ(foo(5)))` rewritten to `foo(6)`:
```
Rule 2688 [label: N/A] [loc: 8,6] [substs: 3]
  Var'Unds'DotVar0 = [kore]
  Var'Unds'DotVar1 = [kore]
  VarX = [kore]
Hook: INT.add @ 0:0:0:0:0:0
  Function: Lbl'UndsPlus'Int'Unds'{} @ 0:0:0:0:0:0
    Term: [kore]
    Term: [kore]
  Result: [kore]
Config: [kore]
```

# The working branch

The work for subtasks marked finished above is in the `builtin-hooks` branch.

IljaZakharov:
This closes #459 except there is still impossible to run K files due to substitutions.

Depends on (and includes) changes from PR #683.


generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -190,7 +190,7 @@ def enter_context(self, location):
         # Check that whether it is the first simplification or not
         if not self._simplification_stack:
             config_with_hole, sub_pattern, location_remained = self._subpattern_search_rec(
-                location, self._current_config, HOLE
+                location[:-1], self._current_config, HOLE
```
IljaZakharov, line 193: I played with real examples and realized that this is necessary but only for the first location for a simplification batch.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -203,7 +203,12 @@ def enter_context(self, location):
         new_info = SimplificationInfo(location, sub_pattern, sub_pattern, 0, self.prover.trivial_proof(sub_pattern))
         self._simplification_stack.append(new_info)
 
-    def apply_hook_simplification(self, symbol: Symbol, substitution: dict[int, Pattern], result: Pattern) -> None:
+    def apply_hook_simplification(self, substitution: dict[int, Pattern], result: Pattern) -> None:
+        # Get the hooked symbol
+        simplified_pattern = self._simplification_stack[-1].initial_pattern
+        symbol, _ = kl.deconstruct_nary_application(simplified_pattern)
+        assert isinstance(symbol, Symbol), f'Pattern {simplified_pattern} is not supported'
+
```
IljaZakharov, line 211: Do we also check that the symbol has hooked attribute?

generation/src/tests/integration/rewrite/test_llvm_proof_generator.py
```diff
@@ -193,3 +193,38 @@ def test_proof_trace_tree_reverse_without_int() -> None:
...
+
+    # First rewrite
+    hint = next(iterator, None)
+    assert isinstance(hint, RewriteStepExpression)
+    assert hint.axiom.ordinal == 2688
+    assert len(hint.substitutions) == 3
+
+    # Function Event
+    hint = next(iterator, None)
+    assert isinstance(hint, FunEvent)
```
fiedlr, line 219: First comes `FunEvent`? How does it know that it is a hint event without rule ordinal?

fiedlr: Aha I see.. I remember that `RuleEvent` is triggered too for these cases of `FunEvent`.

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
```diff
@@ -77,10 +78,37 @@ def _iterate_over_hints(
...
+                yield FunEvent(name, fun_simplification_location)
+            case LLVMHookEvent(name, _, args, result):
+                result_pattern = language_semantics.convert_pattern(result)
+
+                # Processing each argument would require computing results for each argument, which is not supported yet.
+                # Instead we taking the first simplification argument to compute the LHS of the whitelist assumption and
+                # use the result of the hook as the RHS of the whitelist assumption.
+                assert (
+                    len(args) >= 1
+                ), f'Expect at least one simplification in the hook event {name} but got {len(args)}'
```
fiedlr, line 92: Wll this not be always `but got 0`?

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
```diff
@@ -77,10 +78,37 @@ def _iterate_over_hints(
                 non_empty = True
                 yield RewriteStepExpression(axiom, frozendict(converted_substitutions))
             case LLVMFunctionEvent(name, location_str, _):
-                location: tuple[int, ...] = tuple(map(int, location_str.split(':')))
+                fun_simplification_location: tuple[int, ...] = _convert_location(location_str)
                 non_empty = True
-                yield FunEvent(name, location)
+                yield FunEvent(name, fun_simplification_location)
+            case LLVMHookEvent(name, _, args, result):
```
fiedlr, line 84: Should we not call these `args` here `child_fun_events` then?

generation/src/proof_generation/k/execution_proof_generation.py
fiedlr: I guess we could for due diligence, but it is not high priority... You can always check what assumptions were added.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -400,13 +405,23 @@ def rewrite_event(self, rule: KRewritingRule, substitution: dict[int, Pattern])
         return proof
 
     def function_simplification_event(self, location: Location) -> None:
```
fiedlr, line 407: Why have this function?

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -480,6 +495,8 @@ def from_proof_hints(
                         raise NotImplementedError(f'Unsupported axiom type: {str(axiom)}')
                 case FunEvent(_, position):
                     proof_expr.function_simplification_event(position)
+                case HookEvent(_, arg_patterns, result_pattern):
+                    proof_expr.hook_simplification_event(arg_patterns, result_pattern)
```
fiedlr, line 499: This looks like `hook_simplification_event` was the analogue of `function_simplification_event`, but not sure we can do much about it at this point.

fiedlr (approved):
Looks great, Ilja, thank you. I just left a couple of minor comments.

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -169,7 +190,7 @@ def enter_context(self, location):
         # Check that whether it is the first simplification or not
         if not self._simplification_stack:
             config_with_hole, sub_pattern, location_remained = self._subpattern_search_rec(
-                location, self._current_config, HOLE
+                location[:-1], self._current_config, HOLE
```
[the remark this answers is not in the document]
IljaZakharov: @fiedlr It affects any location for simplification that points to the subterm in the *configuration*. All other relative locations are processed as before and we shouldn't change them. But the very first location of each simplification should be truncated.  My guess is that it depends on how the backend counts symbols from the generated top.

generation/src/proof_generation/k/execution_proof_generation.py
IljaZakharov: I'll add the TODO. We don't have the `hooked` attribute in `KSymbol` class. We can add it as a separate PR and implement the assertion mentioned above.

generation/src/proof_generation/k/execution_proof_generation.py
IljaZakharov: @fiedlr The `FunEvent` just has the location and then we can process either hook events and rule simplification events.

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
IljaZakharov: Applied, thanks!

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
IljaZakharov: Right :)

generation/src/tests/integration/rewrite/test_llvm_proof_generator.py
IljaZakharov: Yeah, the `FunEvent` has only the location.

PR #701, opened 2024-01-11
Instantiate dneg_intro after proving it for phi0

MirceaS:
This change reduces the size of the `double-rewrite` proof in the `funct` branch from 24,350 bytes to 3,386 bytes.
We can probably get more gains by using this trick elsewhere also but for now this is a very significant improvement.


fiedlr (approved):
Looks good. 

This shows that we should avoid parametrization of proofs as much as we can, because it breaks memoization of those subproofs.

PR #702, opened 2024-01-11
Basic test harness for circom files

### LINKED ISSUE #697: Add basic testing infrastructure for Circom
Circom doesn't offer much for testing circuits. For starters, we'd like to have a Python testing harness which:
- Contains the actual tests in Python-files (in some reasonable format; in the beginning it can be a tuple of parameters and a reference towards a json file).
- Automates generating the correct circuit for each test.
- Automates computing the witness for a given test and assessing the expected outcome.

traiansf:
Fixes #697


circom/README.md
```diff
@@ -0,0 +1,42 @@
+# Circuits for Matching Logic
+
+## Testing
+
+`circom_testing_harness.py` provides basic infrasctructure for executing tests.
```
mcalancea, line 5: infrastructure

circom/README.md
```diff
@@ -0,0 +1,42 @@
...
+It can either be used as a library to build new tests or can be run directly to run tests described in a JSON input file.
+
+### Running it direclty
```
mcalancea, line 8: directly

circom/README.md
```diff
@@ -0,0 +1,42 @@
...
+- `--test_set_file` to indicate the JSON file containg the description of tests to be performed (defaults to `tests/test_set.json`)
+- `--clean` to specify that no tests should be run, but the environment corresponding to the test-set file should be cleaned.
+
+### Using it as a library
+
+`circom_testing_harness.py` provides the following functionality:
+
+- a class `CircomTest` providing a `run()` method for running a test and a `clean()` method for cleaning its generated files
+  
+  `CircomTest` has the following constructors:
```
mcalancea, line 21: I think this description is a bit too low-level for the README, it can go into the actual code. I would expect the README to describe the purpose of each file and to list useful commands explicitly.

circom/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+    parser.addoption(
+        "--test_set_file", action="store", default='tests/test_set.json', help="location of test-set JSON file"
+    )
+    parser.addoption(
+        '--clean', action='store_true', default=False,
+        help="Don't run tests. Just clean the files generated by the testing process"
+    )
+
+@pytest.fixture
+def clean(request):
```
mcalancea, line 14: I see `--clean` isn't run by default when executing `pytest`. 
- Is this for caching purposes? 
- If I modify the circom source and run `pytest` while forgetting to `--clean` previous runs, am I still testing the old version? If yes, it's too confusion-prone, let's run clean by default and reconsider when the test-suite becomes too slow.

circom/circom_test_harness.py
```diff
@@ -0,0 +1,137 @@
...
+        print("Tested {0} files. Passed {1}. Failed {2}.".format(self.passed + self.failed, self.passed, self.failed));
+        if self.failed_tests:
+            print('Failed Tests:\n\t' + '\n\t'.join(map(lambda t : t.name, self.failed_tests)))
+
+def load_tests(file_name : str):
+    with open(file_name) as file:
+        d = json.load(file)
+        for component in d["components"]:
+            component_name = component["name"]
+            component_lib = component["lib"]
```
mcalancea, line 82: Why?

circom/circom_test_harness.py
```diff
@@ -0,0 +1,137 @@
...
+            print("Component: " + component_lib + '.' + component_name)
+            for instance in component["instances"]:
+                component_params = instance["params"]
+                for test in instance["tests"]:
+                    test_name = test["name"]
+                    test_input = test["input"]
+                    test_output = test["output"]
+                    yield CircomTest(component_name,component_lib, component_params, test_name, test_input, test_output)
+
+def run_tests(tests):
```
mcalancea, line 92: We should probably have only one way of doing things (in this case `pytest`). I'd remove these overlapping features.

circom/circom_test_harness.py
```diff
@@ -0,0 +1,137 @@
...
+        self.name = name
+        self.lib = lib
+
+class CircomComponentInstance(CircomComponent):
+    def __init__(self, name: str, lib: str, params: list[int]) -> None:
+        super().__init__(name, lib)
+        self.params = params
+    
+    def get_runner(self) -> str:
+        return self.lib + '.' + self.name + '_' + '_'.join(map(str, self.params)) + '_runner'
```
mcalancea, line 18:
```suggestion
        return f'{self.lib}.{self.name}_{"_".join(map(str, self.params))}_runner'
```
I think f-strings look more readable; applicable to other lines as well.

mcalancea (changes requested):
Looks good overall, Traian, thank you!
- If lack of default `--clean` can lead to bad caching as described in my comment, I really want it changed
- The README is too verbose for me.
- There are some style comments, it's up to you if you want to fix them. Some might be enforced once we acquire the global poetry env, but I'm not sure which.

mcalancea (approved):
Great! Thank you for the changes!

PR #703, opened 2024-01-11
Cleaning up unit tests for language semantics and proof object generator

IljaZakharov:
There are many small optimizations that can be done to make this tests look better:

- [x] Reuse code for defining semantics with cells
- [x] Add functions for one-line retrieval of sorts and symbols
- [x] Merge two functions for constructing configurations
- [x] Move the test related to adding functional axioms to the proof generation test file


generation/src/tests/unit/test_kore_language_semantics.py
```diff
@@ -29,116 +29,137 @@
...
+
+    with semantics as sem:
+        module = sem.get_module('node-tree')
+        with module as mod:
+            tree_ksort = mod.sort('SortTree')
+            tree_sort = tree_ksort.aml_symbol
+            tree_top = kore_top(tree_sort)
+
+            init_ksymbol = mod.symbol('init', tree_ksort, is_functional=True, is_ctor=True).app
+            next_ksymbol = mod.symbol('next', tree_ksort, is_functional=True, is_ctor=True).app
```
malturki, line 136: This is probably not very important, but are the `init` and `next` symbols intended to be functional in this example?

malturki (approved):
Looks good. I had one tiny question.

PR #708, opened 2024-01-12
Add an example proof decomposed into subproofs

malturki:
This PR adds a simple example proof in propositional logic decomposed into subproofs. It stimulates the following decomposition of the AML claim `{forall n . f(n) => f(n+1)} |- f(0) => f(3)`:

```
{f(0) => f(1), f(1) => f(3)} |- f(0) => f(3) 
{f(1) => f(2), f(2) => f(3)} |- f(1) => f(3) 
{forall n . f(n) => f(n+1)} |- f(0) => f(1) 
{forall n . f(n) => f(n+1)} |- f(1) => f(2) 
{forall n . f(n) => f(n+1)} |- f(2) => f(3)
```


malturki:
> LGTME. I understand that this is just one script that generates the particular example suibproofs. The full, automatic subproof generator is yet to be implemented. Is that right?

That's correct. The automatic subproof generator needs further investigation, and hence this example so that the SDC development is not blocked.

xc93 (approved):

PR #711, opened 2024-01-12
Implement apply_esbust and apply_ssubst

### LINKED ISSUE #319: Implement ESubst/SSubst in zkLLVM checker
- [x] `apply_esubst` (not used in `imp-reflexivity`)
- [x] `apply_ssubst` (not used in `imp-refkexivity`)

yzhang90:
Closes #319


cpp/tests/unit_tests.hpp
```diff
@@ -530,7 +530,7 @@ void execute_vector(std::array<int, MAX_SIZE> &instrs, Pattern::Stack &stack,
   Pattern::execute_instructions(instrs, stack, memory, claims, phase);
 }
 
-void test_publish() {
+int test_publish() {
```
Robertorosmaninho, line 533: Is there any particular reason to change it?

Robertorosmaninho (approved):
Assuming that you have tested at least with `impreflex.inp` and it doesn't break anything.
LGTM!

cpp/tests/unit_tests.hpp
yzhang90: I noticed some tests return int while others return void. What is the difference?

Robertorosmaninho: If I remember correctly, the only advantage to having a test return a value instead of `void` was to check its result with `assigner --check.` Nevertheless, for the final implementation, they probably should all be void as it is on the reference rust implementation.

PR #715, opened 2024-01-12
Update risc0 version

dwightguth:
This updates the version of risc0 used by the tool to 0.19.1. We do this in order that we be able to use the latest version of the prrofiling code, which is much better documented than the version present in 0.16.1.


risc0/pi2/src/main.rs
```diff
@@ -36,26 +36,17 @@ fn main() {
         .unwrap();
 
     // Next, we make an executor, loading the (renamed) ELF binary.
-    let mut exec = default_executor_from_elf(env, GUEST_ELF).unwrap();
+    let prover = default_prover();
 
-    println!("Checking the proof...");
+    println!("Checking the proof and generating the receipt...");
 
     // Run the executor to produce a session.
```
Robertorosmaninho, line 43: We need to update this comment as we're not creating a session anymore.

risc0/pi2/src/main.rs
```diff
@@ -36,26 +36,17 @@ fn main() {
...
-    let session = exec.run().unwrap();
+    let receipt = prover.prove_elf(env, GUEST_ELF).unwrap();
 
     let runtime = now.elapsed().as_millis();
 
     println!("Ran in {} ms", runtime);
```
Robertorosmaninho, line 48: We need to update this log message as we're now already proving it here instead of only executing then proving it.

risc0/pi2/src/main.rs
```diff
@@ -36,26 +36,17 @@ fn main() {
...
 
     // Next, we make an executor, loading the (renamed) ELF binary.
-    let mut exec = default_executor_from_elf(env, GUEST_ELF).unwrap();
+    let prover = default_prover();
 
-    println!("Checking the proof...");
+    println!("Checking the proof and generating the receipt...");
 
     // Run the executor to produce a session.
-    let session = exec.run().unwrap();
```
Robertorosmaninho, line 44: I'm assuming we're no longer able to execute the program without proving it, right? Is this no longer supported by RISC0? (I think so, but just double-check.) 

This is important because if that's the case, we'll need to delete a column from [`Performance.md`](https://github.com/runtimeverification/proof-checker/blob/main/Performance.md#risc-zero-v0161) that is the "CPU/GPU Exec Time"

risc0/pi2/src/main.rs
```diff
@@ -36,26 +36,17 @@ fn main() {
...
+    let receipt = prover.prove_elf(env, GUEST_ELF).unwrap();
 
     let runtime = now.elapsed().as_millis();
```
Robertorosmaninho, line 46: Change the name to "provetime" to be more significative.

risc0/tests/direct-implementation/src/main.rs
```diff
@@ -16,40 +16,28 @@ fn main() {
     let env = ExecutorEnv::builder().build().unwrap();
 
     // Next, we make an executor, loading the (renamed) ELF binary.
-    let mut exec = default_executor_from_elf(env, GUEST_CSL_ELF).unwrap();
+    let prover = default_prover();
 
-    println!("Checking the proof...");
+    println!("Checking the proof and generating the receipt...");
 
     // Run the executor to produce a session.
```
Robertorosmaninho, line 23: "Run the prover"

risc0/tests/direct-implementation/src/main.rs
```diff
@@ -16,40 +16,28 @@ fn main() {
...
-    let session = exec.run().unwrap();
+    let receipt = prover.prove_elf(env, GUEST_CSL_ELF).unwrap();
 
     let runtime = now.elapsed().as_millis();
```
Robertorosmaninho, line 26: provetime

risc0/tests/direct-implementation/src/main.rs
```diff
@@ -16,40 +16,28 @@ fn main() {
...
 
     println!("Ran in {} ms", runtime);
```
Robertorosmaninho, line 28: `"Proved in {} ms"`

Robertorosmaninho (approved):
I asked for some simple changes; the main thing here is just to double-check that we can't execute it anymore without proving it to avoid deleting information from the Performance table. 

Other than that, the modification LGTM!
Thanks for taking time to update this!

risc0/pi2/src/main.rs
dwightguth: You can still execute the program without proving it, but there doesn't seem to be an API for executing the program without proving it, then proving the executed trace.

Robertorosmaninho: Got it, so unfortunately, if we can't split the operations, I think we'll be forced to remove this column from the table.

Robertorosmaninho (approved):
LGTM

PR #718, opened 2024-01-15
Finish a working E2E example with function events

fiedlr:
Adds a first working E2E example with function events to the repo (`simplify.tree-reverse`)


generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
```diff
@@ -109,4 +109,4 @@ def _iterate_over_hints(
 
 
 def _convert_location(location: str) -> tuple[int, ...]:
-    return tuple(map(int, location.split(':')))
+    return tuple(map(int, location.split(':')))[1:]
```
IljaZakharov, line 112: I didn't get that. Making this change here means that all locations from hints should be truncated. But relative locations are fine. My understanding is that only the first location that points to a subpattern inside `kcell` is too long. But maybe I am missing something.

IljaZakharov (approved):
Looks nice! I approved the changes, but I have concerns regarding the change in the `rewite_steps`. I tried exactly the same while testing the `dv.k` example, and it didn't work because all relative locations become truncated. But, again, maybe I am missing something.

generation/src/proof_generation/k/kore_convertion/rewrite_steps.py
fiedlr: There's an extra `0` in the pretty printed proof hint that's needed for the algorithm (also in nested events).

Locations in tests might be skewed because we are throwing out injections

fiedlr: A follow-up PR will be to make `dv` work, we'll see there

Issue #685, opened 2024-01-09
Add easy-to-use pi2 scripts in Python


fiedlr:
- [ ] We need some easy-to-use scripts for each one of the tests described in https://runtimeverification.slack.com/archives/C05MF3KEWNM/p1704736400610179

Note that we already have a script for the `gen-hint` step here: https://github.com/runtimeverification/proof-checker/blob/main/generation/scripts/gen-execution-proof-hints.sh

- [ ]  We need a single script called `pi2` that will call all the scripts from Step 1 in sequence. I started some work on it in this PR: https://github.com/runtimeverification/proof-checker/pull/390/files , but it is likely a bit  outdated

Robertorosmaninho:
- create this in the `generation` subdirectory

Robertorosmaninho:
Goal Infrastructure:

```bash
kompile <definition.k>
gen-hint <kompiled-dir> <input.kore> -o <hint>
gen-proof <kompiled_dir> <hint> -o <proof>
check-proof <proof>
```

Robertorosmaninho:
# Roadmap

## Basic Structure
- [ ] Create the poetry project
- [ ] Create the main script `pi2`
- [ ] Create the Check Proof subproject
- [ ] Create the Gen Proof subproject
- [ ] Create the Gen Hints subproject
- [ ] If necessary, create the kompile subproject

## Check Proof
- Given the `<proof>` the script must generate the appropriate input file or program with the input hardcoded.
- The output of this script must be `[PROVED AND VERIFIED SUCCESSFULLY]` or `[FAILED TO PROVE OR VERIFY]: ERROR_MESSAGE`
- [ ] Modify the benchmark scripts to use a standalone `build.sh`
- [ ] Rename the build scripts to `check_with_<zk_backend>.sh`. E.g. `check_with_cairo.sh`.
- [ ] Modify the build scripts to have an option to check if the backend is available. E.g. <zk-tool> --version works as expected.
- [ ] On the Check Proof file add options to check which zk backends are availables -> check if the all build scripts exists and the return of `check_with_<zk_backend>.sh --version` doesn't return an error.
- [ ] Create infrastructure necessary to select the zk backend from the CLI and all things needed to use it.
- [ ] Create an option `--debug` that present the times used to benchmarking the backends.

PR #720, opened 2024-01-15
Introducing `check_with_risc0.sh`

Robertorosmaninho:
This PR splits the old `bench.sh` script into a new `bench.sh` that just calls the new `check_with_risc0.sh` script for every benchmark.

This new `check_with_risc0.sh` is a new user-friendly CLI tool to execute our proof-checker using risc0 as zk-backend and our binary proofs! 

This is part of #685 effort to generate the `check-proof` program.

```bash
# Usage: check_with_risc0.sh [options] <input directory>
#
# Options:
#  -a, --available            Print if the program is available or not
#  -h, --help                 Print this help message
#  -o, --output FILENAME      Output file
#  -s, --stats                Print statistics about the input
```


yzhang90 (approved):
LGTM

PR #722, opened 2024-01-15
Implement the can_be_replaced by function for MetaVars and associated helpers

MirceaS:


generation/src/proof_generation/aml/syntax.py
```diff
@@ -118,6 +142,18 @@ class SVar(Pattern):
     def evar_is_fresh_ignoring_metavars(self, name: int, ignored_metavars: frozenset[int]) -> bool:
         return True
 
+    def svar_is_fresh(self, name: int) -> bool:
```
IljaZakharov, line 145: Why is it not implemented?

fiedlr: We agreed not to implement any functions that are not currently needed.

generation/src/proof_generation/aml/syntax.py
```diff
@@ -157,6 +193,18 @@ class Symbol(Pattern):
...
```
fiedlr, line 196: We cannot have it implemented here and not in the `EVar`, for example. Either implement it nowhere or everywhere that it can be implemented without the `svar_is_app_ctx_hole`

generation/src/proof_generation/aml/syntax.py
```diff
@@ -157,6 +193,18 @@ class Symbol(Pattern):
...
+        return True
```
fiedlr, line 197:
```suggestion
        raise NotImplementedError
```

fiedlr (changes requested):
Still did not finish, submitting what I have

malturki (approved):
Looks good to me

generation/src/proof_generation/aml/syntax.py
fiedlr: I guess we can remove the overriding occurrences too?

generation/src/proof_generation/aml/syntax.py
```diff
@@ -235,6 +295,20 @@ def evar_is_fresh_ignoring_metavars(self, name: int, ignored_metavars: frozenset
...
+
+    def svar_is_positive(self, name: int) -> bool:
+        return self.left.svar_is_positive(name) and self.right.svar_is_positive(name)
+
+    def svar_is_negative(self, name: int) -> bool:
+        return self.left.svar_is_negative(name) and self.right.svar_is_negative(name)
+
+    def evar_is_app_ctx_hole(self, name: int) -> bool:
+        return (self.left.evar_is_app_ctx_hole(name) and self.right.evar_is_fresh(name)) or (
+            self.left.evar_is_fresh(name) and self.right.evar_is_app_ctx_hole(name)
```
fiedlr, line 309: One case is missing, where it's fresh on rhs and app_ctx on pjs

generation/src/proof_generation/aml/syntax.py
```diff
@@ -235,6 +295,20 @@ def evar_is_fresh_ignoring_metavars(self, name: int, ignored_metavars: frozenset
...
+        raise NotImplementedError
+
+    def svar_is_positive(self, name: int) -> bool:
+        return self.left.svar_is_positive(name) and self.right.svar_is_positive(name)
+
+    def svar_is_negative(self, name: int) -> bool:
+        return self.left.svar_is_negative(name) and self.right.svar_is_negative(name)
+
+    def evar_is_app_ctx_hole(self, name: int) -> bool:
+        return (self.left.evar_is_app_ctx_hole(name) and self.right.evar_is_fresh(name)) or (
```
fiedlr, line 308: This line is incorrect. It would return true for two occurrences of var, but the occurrence has to be exactly one

generation/src/proof_generation/aml/syntax.py
MirceaS: I don't think we should remove those. They stand there as a placeholder for the work that is to come.

fiedlr (changes requested):

generation/src/proof_generation/aml/syntax.py
fiedlr: Oops, I read it wrong. Sorry

fiedlr (approved):
Looks great! Thanks.

PR #724, opened 2024-01-15
Modify `lurk` file structure

Robertorosmaninho:
This PR reacts to the new lurk feature https://github.com/lurk-lab/lurk-rs/pull/1040 that fixes the `!load` meta-command, enabling it to use relative paths to import definitions from other files. 

After this PR, instead of having all files in the same directory, we'll have a nice organization as follows:

```bash
- lurk
  |_ src
  |_ tests
      |_ direct-implementation
      |_ proofs-of-proofs
      |_ unit-tests
  scripts
```


yzhang90 (approved):
LGTM

Issue #578, opened 2023-12-06
Update checkers based on the updates in the RISC0 checker


fiedlr:
We made the following changes in the checker, please sync with the rest of the implementations:

1. https://github.com/runtimeverification/proof-checker/pull/555
2. https://github.com/runtimeverification/proof-checker/pull/568

Robertorosmaninho:
Reopening it as we still need to downstream the modifications to LURK

PR #726, opened 2024-01-16
Update zkllvm implementation to be consistent with the rust reference implementation

yzhang90:
zkllvm part of #578


Robertorosmaninho (approved):
LGTM

PR #730, opened 2024-01-16
Print stack info with claim and proof publishing

fiedlr:
Updates the pretty printer to print stack info for each published claim and proof.


generation/src/proof_generation/interpreter/pretty_printing_interpreter.py
```diff
@@ -178,6 +178,7 @@ def load(self, id: str, term: Pattern | Proved) -> None:
     @pretty(print_stack=False)
     def publish_proof(self, proved: Proved) -> None:
         self.out.write('Publish')
+        self._print_stack()
```
Robertorosmaninho, line 181: Why call `_print_stack` here and below instead of `print_stack`?

fiedlr: Because I want to ignore the flag, so I just extracted the functionality I wanted

fiedlr: `print_stack()` is still used by the @pretty decorator. Basically if `pretty` flag is turned on, the stack should be  printed with every stack change

Robertorosmaninho (approved):
LGTM!

generation/src/proof_generation/interpreter/pretty_printing_interpreter.py
Robertorosmaninho: Got it! I just wanted to make sure that this was intentional!

proofs/generated-from-k/double-rewrite/foo-a.double-rewrite.pretty-claim
```diff
@@ -277,4 +279,7 @@ Instantiate 0, 1
...
+	1: (<ksym_Lbl'-LT-'generatedTop'-GT-'> <ksym_Lbl'-LT-'k'-GT-'> (ksym_inj(ksort_SortFoo, ksort_SortKItem, ksym_LblFooA'LParRParUnds'DOUBLE-REWRITE-SYNTAX'Unds'Foo()) ~> ksym_dotk()) </ksym_Lbl'-LT-'k'-GT-'> <ksym_Lbl'-LT-'generatedCounter'-GT-'> dv(0):ksort_SortInt </ksym_Lbl'-LT-'generatedCounter'-GT-'> </ksym_Lbl'-LT-'generatedTop'-GT-'> k=> <ksym_Lbl'-LT-'generatedTop'-GT-'> <ksym_Lbl'-LT-'k'-GT-'> (ksym_inj(ksort_SortFoo, ksort_SortKItem, ksym_LblFooB'LParRParUnds'DOUBLE-REWRITE-SYNTAX'Unds'Foo()) ~> ksym_dotk()) </ksym_Lbl'-LT-'k'-GT-'> <ksym_Lbl'-LT-'generatedCounter'-GT-'> dv(0):ksort_SortInt </ksym_Lbl'-LT-'generatedCounter'-GT-'> </ksym_Lbl'-LT-'generatedTop'-GT-'>):ksort_SortGeneratedTopCell
```
MirceaS, line 284: This implementation results in the Stack being printed on the same line as the `Publish` command, which is awkward. It also has the side effect that if we do enable stack printing, this stack for the `Publish` will be printed twice.
I think we should instead implement this by somehow figuring out that the instruction being processed in line 41 is a `Publish` and using this boolean information in the if on line 44 (of the pretty printing interpreter file).
We could have each function return this information as a boolean, since at the moment they're all just returning `None` anyway.

MirceaS (changes requested):

proofs/generated-from-k/double-rewrite/foo-a.double-rewrite.pretty-claim
fiedlr: Not sure why it's awkward. Publish finishes a proof block. I want to see what state stack is in after finishing a proof block.

fiedlr: `print_stack` is a flag determining if `print_stack()` is even called. Then `print_stack()` checks if the flag is turned on.

MirceaS: `Publish Stack` sounds like a command that publishes the stack. It's confusing for a reader to see them both on the same line.
If we have that decorator, I propose the following: The decorator takes one more boolean argument: `force_print_stack`, set to `False` by default, which, when active, will print the stack regardless of the flag in the `pretty_options`. This should be easy to implement by having the `print_stack` take this `force_print_stack` argument as input.

MirceaS: This also means that we don't need to have two separate `print_stack` functions.

fiedlr: OK

generation/src/proof_generation/interpreter/pretty_printing_interpreter.py
```diff
@@ -178,26 +178,21 @@ def load(self, id: str, term: Pattern | Proved) -> None:
     @pretty(print_stack=False)
```
MirceaS: modify this as well.
Also do we also want the stack printed for `publish_axiom`? Since that wasn`t there in the first iteration.

fiedlr: That was a bug, fixed

proofs/generated-from-k/double-rewrite/foo-a.double-rewrite.pretty-claim
fiedlr: How about now?

MirceaS:
With the risk of sounding a bit pedantic, I wouldn't name the new variable `ignore_pretty_opts` because to me that suggests that we're ignoring what the opts say and instead just fall back on what the other decorator argument says, which is not what's happening. The `ignore` variable forces the stack to be printed no matter what. But I'll leave this decision up to you.

On that note however, I would change the decorator calls so that if we set the `ignore` flag, we don't also set the original flag, since it's irellevant, and it could be confusing to have them both (since the `ignore` one takes precedence).

MirceaS:
> With the risk of sounding a bit pedantic, I wouldn't name the new variable `ignore_stack_opts` because to me that suggests that we're ignoring what the opts say and instead just fall back on what the other decorator argument says, which is not what's happening. The `ignore` variable forces the stack to be printed no matter what. But I'll leave this decision up to you.
> 
> On that note however, I would change the decorator calls so that if we set the `ignore` flag, we don't also set the original flag, since it's irellevant, and it could be confusing to have them both (since the `ignore` one takes precedence).

Alternatively, you could change the logic so that the variable does what I said in my interpretation, which would also allow us to never print the stack for certain commands, even if the user enables the stack printing option.

malturki (commented):
Why not just set `print_stack=True` for `publish_proof` and `publish_claim` (without adding any extra flags)? Wouldn't this achieve the intended behavior?

fiedlr:
> Why not just set `print_stack=True` for `publish_proof` and `publish_claim` (without adding any extra flags)? Wouldn't this achieve the intended behavior?

No, because we have a CLI flag we want to bypass. Only setting it to true would not show the stack for publish with the flag turned on

fiedlr:
> With the risk of sounding a bit pedantic, I wouldn't name the new variable `ignore_stack_opts` because to me that suggests that we're ignoring what the opts say and instead just fall back on what the other decorator argument says, which is not what's happening. The `ignore` variable forces the stack to be printed no matter what. But I'll leave this decision up to you.
> 
> 
> 
> On that note however, I would change the decorator calls so that if we set the `ignore` flag, we don't also set the original flag, since it's irellevant, and it could be confusing to have them both (since the `ignore` one takes precedence).

I'll rename it to call_print_stack and force_stack

For the ordering thing, that's not really true.

Simply having force_stack=True does not show the stack because the print function isn't called

MirceaS (approved):

PR #731, opened 2024-01-16
Fixing Performance.md formating

Robertorosmaninho:
This only fix formatting for dev and the labels on lurk that where all GPU instead of CPU and GPU


IljaZakharov (approved):
Looks good! But there are still a couple of things to improve:
- No links to benchmarks in the RISC Zero (v0.16.1) table (comparing with the other tables)
- It would be nice to have the same set of columns between tables (exactly same names and order). How does it sound to add extra comments and give new columns some highlighting in cases when it is not possible.

PR #739, opened 2024-01-17
Support Risc0 profiler

dwightguth:
We remove the profiler for the rust code that profiles natively and replace it with a profiler that profiles the number of risc-v cycles used by the risc0 guest. This is a much more accurate measure of performance.


fiedlr (approved):
Looks great! Before merging, please just fix the typo I noticed and update the README (updating the old profiling instructions).

PR #741, opened 2024-01-17
Replace `Rc` with leaky pointers in RISC0 Rust code

dwightguth:
This PR replaces reference counted pointers everywhere in the Rust proof checker with a wrapped `Ptr` class which simply wraps a raw rust pointer and provides some trait and function implementations around it. The memory allocated by these pointers is allocated via `Box::new` and then converted to raw pointers via `Box::leak`. As a result, this memory is never freed.

Normally this would be a poor development choice because of the memory leaks leading to memory never being freed across the lifetime of an application. However, the memory actually used inside the RISC0 guest is very small compared to the amount of memory used by the ZKVM. Furthermore, the time spent proving the program so far outstrips the time spent executing the program that we see significant performance improvements even just by skipping deallocation of memory when it is no longer reachable. While this would in theory be a scalability concern, the nature of proofs in Pi^2 is that they are composable; we will instead choose to execute a number of smaller ZK proofs rather than one bigger one, which means that the memory used by each individual proof during execution will never grow to the point where it becomes an issue that it is never deallocated during execution.


nishantjr:
@dwightguth I don't like the way that `unsafe` is sprinkled everywhere. This seems to allow all sorts of unsafe behaviour.  It there someway we can isolate it to `Ptr` ?

nishantjr:
@dwightguth Ideally, if it's not too much work it'd be nice to be able to hide the choice of pointer implementation behind a typedef. That way, we can use the non-leaky version if we don't need ZK.

bmmoore:
It looks like the only operation actually requiring `unsafe` is a pointer dereference in `Ptr`, and that seems to be safely encapsulated. Leaking per-se is not unsafe behavior, so there is no need to tag all the functions using `Ptr` as unsafe. (Also, Ptr could probably be implemented without any unsafe code by instead holding a `&'static T` produced at construction time by `Box::leak` on a new `Box`).

dwightguth:
@nishantjr I'm not sure it makes sense to try to make the type swappable. Right now the code I wrote is heavily dependent on the fact that `Ptr` implements `Copy` and `Deref` and that is a choice that is not necessarily going to be implemented by all possible pointer types we could deal with. For example, `Rc` does not implement `Copy`. My understanding of the logic here is that we're going to be operating on small enough examples that the leak doesn't matter; that's true whether we are using ZK or not. Is that understanding incorrect?

risc0/pi2/guest/src/main.rs
```diff
@@ -34,7 +34,9 @@ pub fn main() {
     #[cfg(debug_assertions)] {
         _begin = env::get_cycle_count();
     }
-    verify(gamma_buffer, claims_buffer, proof_buffer);
+    unsafe {
```
nishantjr, line 37: This shouldn't be needed now.

nishantjr:
That makes sense @dwightguth, its not worth it.

nishantjr (approved):
Looks good!

nishantjr:
> (Also, Ptr could probably be implemented without any unsafe code by instead holding a `&'static T` produced at construction time by `Box::leak` on a new `Box`)

Oh that's interesting @bmmoore . But how can we construct an object with static lifetime, without using a global constant?

PR #746, opened 2024-01-18
Introducing pi2 script

### LINKED ISSUE #685: Add easy-to-use pi2 scripts in Python

Robertorosmaninho:
Closes #685 

Check the issue related to see a full description of this task.

- [x] pi2
- [x] Check-Proof Script
- [x] Gen-Proof Script
- [x] Gen-Hints Script
- [x] Kompile Script


pi2/README.md
```diff
@@ -0,0 +1,24 @@
...
+poetry shell
+```
+
+# Run the scripts
+```bash
+# Main script
+pi2 <definition.k> <input.kore> [OPTIONS]
+
+# Sub Scripts
+kompile <definition.k>
```
IljaZakharov, line 20: Personally, I don't like the fact that scripts pollute the current directory by default. We have places for both definition and hints in `.build`. I find it useful because one always knows where to find files, and they don't appear in git.

pi2/README.md
```diff
@@ -0,0 +1,24 @@
...
+gen-hint <kompiled-dir> <input.kore> -o <hint>
```
IljaZakharov, line 21: It also returned nothing for the `--help` argument.

pi2/README.md
```diff
@@ -0,0 +1,24 @@
...
+gen-proof <kompiled_dir> <hint> -o <proof-directory>
+check-proof <zk-backend> <proof-directory>
```
IljaZakharov, line 23: `check-proof -a` just crashed.

pi2/README.md
```diff
@@ -0,0 +1,24 @@
+# Requirements
+Have a working K installation with the appropriate version.
```
IljaZakharov, line 2: What about ZK backends?

pi2/pi2/check_proof.py
```diff
@@ -0,0 +1,194 @@
...
+    if zk_backend not in supported_backends:
+        click.echo("Unsupported backend: " + zk_backend)
+        return
+    
+    # Check if the provided backend is available
+    if zk_backend not in available_backends:
+        click.echo("Backend not available: " + zk_backend)
+        return
+    
+    script_name = "./check_with_" + zk_backend + ".sh"
```
IljaZakharov, line 121: Let's use `os.path` or `pathlib` for working with files and paths.

pi2/pi2/check_proof.py
```diff
@@ -0,0 +1,194 @@
...
+        click.echo("Missing argument PROOF-DIRECTORY")
+        return
+    
+    # Check if the provided proof directory exists
+    if not translate_input and not translate_input_off and not os.path.isdir(proof_directory):
+        click.echo("Directory not found: " + proof_directory)
+        return
+    
+    # Collect scripts arguments
+    arguments = []
```
IljaZakharov, line 145: Is there any way just to forward these arguments without going through all of them? I know it is possible with `argparse`, but I never worked with `click`.

pi2/pi2/gen_hints.py
```diff
@@ -0,0 +1,78 @@
...
+
+# Usage: gen-hint <KOMPILED-DIR> <INPUT-KORE> [OPTIONS]
+
+# Options:
+#   -h, --help                Show this message and exit.
+#   -o, --output FILENAME     Save the generated hinto into FILENAME. Default is {input}.hints.
+#   -v, --verbose             Show verbose output.
+
+def check_program_exists(name):
+    cmd = "where" if platform.system() == "Windows" else "which"
```
IljaZakharov, line 15: We definitely want to rely on some library for such matters. One of `distutils` or `shutils`.

pi2/pi2/gen_hints.py
```diff
@@ -0,0 +1,78 @@
...
+@click.option("--verbose", "-v", type=bool, default=False, is_flag=True,
+                help="Show verbose output.")
+
+def gen_hints(kompiled_dir: str, input_kore: str, output: str, verbose: bool) -> None:
+    
+    # Get the root directory of the project
+    root_directory = subprocess.check_output(
+        ["git", "rev-parse", "--show-toplevel"]).strip().decode("utf-8")
+    
+    script_name = "gen-execution-proof-hints.sh"
```
IljaZakharov, line 35: Same about working with paths.

pi2/pi2/gen_hints.py
IljaZakharov: And I doubt that we support Windows at all.

IljaZakharov (changes requested):
Looks good, great job! 

But I have a number of suggestions to make it shine:
- It is unclear where to find instructions for installing backends and where is the location of supported K files. Both top-level and pi2 README files require improvements. 
- I would propose using `.build`directory by default if a user doesn't specify any output directories.
- Some scripts unexpectedly crash (see my comments) or I failed to find them in the path at all. 
- Please use our default formatting and type checking for the Python code.
- I left a couple of really minor coding suggestions for working with paths and options.

pi2/pi2/check_proof.py
Robertorosmaninho: It isn't because `click` has its way to parse variables through the CLI, by already setting them as arguments to the main program.

pi2/README.md
Robertorosmaninho: Fixed

pi2/README.md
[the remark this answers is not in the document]
Robertorosmaninho: sorry, the correct command is `gen-hints`

Robertorosmaninho: Please be sure you installed the pi2 project and its using the correct environment either by `poetry shell -C pi2` or `source ${path_of_virtual_env}`
Ex.:
```bash
$ poetry shell -C pi2
Virtual environment already activated: /Users/robertorosmaninho/Library/Caches/pypoetry/virtualenvs/pi2-ubaZfU4E-py3.11
$ gen-hints --help
Usage: gen-hints [OPTIONS] [KOMPILED_DIR] [INPUT_KORE]

Options:
  -o, --output TEXT  Save the generated into into FILENAME. Default is INPUT-
                     KORE.hints.
  -v, --verbose      Show verbose output.
  --help             Show this message and exit.
```

I'll add this information to the `README.md`

pi2/README.md
```diff
@@ -0,0 +1,24 @@
...
+
+# Activate the virtual environment
+```bash
+poetry shell
+```
+
+# Run the scripts
+```bash
+# Main script
+pi2 <definition.k> <input.kore> [OPTIONS]
```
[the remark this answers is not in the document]
Robertorosmaninho: We can call it `program_input.kore` or something like that, just to be clear that:
1. The program is the input for the given definition (workflow and gramma inherit from K)
2. Make it clear that the script and our workflow expect a `KORE` file.

About "where to find K definitions", I can add the path that contains our examples on `README.md`, but the idea is that people will bring "their own" K definition and KORE input.

pi2/pi2/gen_hints.py
Robertorosmaninho: This is just the name of the script, the path that may be an issue is below

pi2/README.md
Robertorosmaninho: I agree with you if we consider ourselves as the main users of this script. However, if we think of a user outside of RV who will use this tool in his/her own workflow, the default output directory being a hidden folder on our directory doesn't seem intuitive.

Robertorosmaninho:
Hey @IljaZakharov, I think I solved most, if not all, of your comments and added the `-d/--default-directory` flag to all scripts so they can use the `.build` directory to output their results!

IljaZakharov (approved):
Looks nice, thank you!

PR #750, opened 2024-01-18
Introduce kore validity

MirceaS:


generation/src/proof_generation/proofs/kore.py
```diff
@@ -37,9 +37,14 @@
...
 
+""" kore_top(sort) """
+kore_top = Notation('kore-top', 1, App(inhabitant_symbol, phi0), 'k⊤:{0}')
+
 """ in_sort(element, sort) """
-in_sort = Notation('in-sort', 2, subset(phi0, App(inhabitant_symbol, phi1)), '{0}:{1}')
+in_sort = Notation('in-sort', 2, subset(phi0, kore_top(phi1)), '{0}:{1}')
 
+""" kore_valid(sort, pattern) """
+kv = Notation('kore-valid', 2, equals(phi1, kore_top(phi0)), '⊢{1}:{0}')
```
fiedlr, line 47: This is also fine but note that in the pretty printer on the proof stack, it's gonna look like this:

```
⊢ ⊢ ({1}:TopCell):TopCell
```

because of the `Proved` pretty printing and the declaration of the outermost sort.
I was trying to avoid that. Can we please use something like this?

```
⊢ ⊩{sort} phi
```

To differentiate between the embedded proving and sort declaration?

fiedlr, line 47:
```suggestion
kv = Notation('kore-valid', 2, equals(phi1, kore_top(phi0)), '⊩{0} {1}')
```

IljaZakharov (approved):
Looks good to me!

PR #754, opened 2024-01-19
Optimize zkllvm input array size

### LINKED ISSUE #753: Optimize zkllvm input array size
Part of #624 

Instead of a big fixed size which fit all the proof, now we use different fixed size for different proof input.

yzhang90:
Closes #753 

./check_with_zkllvm.sh --transpiler --stats ../proofs/translated/impreflex-compressed-goal/
Use transpiler to generate the proof and verify it is deprecated and will be removed in the future. Please consider use the proof-producer instead.
impreflex-compressed-goal circuit generation .. 0.569 s
impreflex-compressed-goal proof+verification .. 292.704 s


zkllvm/check_with_zkllvm.sh
```diff
@@ -226,15 +230,16 @@ if [ "$translate_input" == true ]; then
...
-    # Setting the inputs variable
-    gamma="$input/$input_filename.ml-gamma"
-    claim="$input/$input_filename.ml-claim"
-    proof="$input/$input_filename.ml-proof"
-
     # Execute the translator with the input files and save the output to a temporary file
     tmp_input_file=$(mktemp)
 
-    ./translator.py "$gamma" "$claim" "$proof" > "$tmp_input_file"
+    result=$(python3 translator.py "$input" -o "$tmp_input_file")
```
Robertorosmaninho, line 236: I forgot that the best way to call this script is with the full path, could you fix it for me?
The idea is to have something like `$main_script_path/translator.py`. Check the others `check_with_*.sh` scripts to see it there.

Robertorosmaninho (changes requested):
Just one small important comment, but overall, it LGTM!

Robertorosmaninho (approved):
LGTM

PR #770, opened 2024-01-22
Introducing the `sdc` project

Robertorosmaninho:
### Introducing the SubProof Dependency Checker
This is the first simple string-based version of `sdc`. It checks for dependency by comparing the strings and it checks the dependencies in parallel on the CPU.
#### Future work:
- Main function accepting inputs as string or binaries: TBD
- Execute in parallel on GPU
- Changes to work on the zk-backends Cairo and/or Rust: TBD

Part of #780


yzhang90 (commented):
LGTM.
1) Have you tried if it works with Risc0?
2) Can we have a negative test case?

Robertorosmaninho:
1.Do you mean test a zk sdc version?
2. Ok

yzhang90:
> Do you mean test a zk sdc version?
Yes. if it is not too difficult to test.

sdc/src/lib.rs
```diff
@@ -100,4 +103,49 @@ mod tests {
...
+        let target_gamma = vec!["forall n . f(n) => f(n+1)"];
+        let target_claim = "f(0) => f(3)";
+        let subproofs = vec![
+            (vec!["forall n . f(n) => f(n+1)"], "f(2) => f(3)"),
+            (vec!["forall n . f(n) => f(n+1)"], "f(1) => f(2)"),
+            (vec!["forall n . f(n) => f(n+1)"], "f(0) => f(1)"),
+            (vec!["f(1) => f(2)", "f(2) => f(3)"], "f(1) => f(3)"),
+            (vec!["f(0) => f(1)", "f(1) => f(3)"], "f(0) => f(4)"), // Claim not used
+            (vec!["f(0) => f(1)", "f(1) => f(3)"], "f(0) => f(3)"),
+
```
yzhang90, line 146: remove this blank line.

sdc/src/lib.rs
```diff
@@ -100,4 +103,49 @@ mod tests {
         let dependecy_check = sdc(target_gamma, target_claim.to_string(), subproofs);
         assert_eq!(dependecy_check, true);
     }
+
+    #[test]
```
yzhang90, line 107: Can we add one more test where the subproofs[len-1].2 != target claim?

Robertorosmaninho: Sure

yzhang90 (approved):
Add risc0 as the dependency, otherwise LGTM

PR #773, opened 2024-01-23
Add the remaining kore.mm notation definitions

### LINKED ISSUE #749: Transfer remaining of kore definitions from `kore.mm`

malturki:
This PR:
1. Adds all the remaining kore notation definitions in `kore.mm` to `kore.py` (listed below)
2. Implements `svar_is_fresh(...)` for patterns, which is needed to evaluate these definitions

The definitions this PR adds are as follows:
- [x] kore-mu
- [x] kore-nu
- [x] kore-all-path-next
- [x] kore-eventually
- [x] kore-weak-eventually
- [x] kore-always
- [x] kore-well-founded
- [x] kore-well-founded-alt
- [x] kore-rewrites-star
- [x] kore-rewrites-plus
- [x] kore-one-path-reaches-star
- [x] kore-one-path-reaches-plus
- [x] kore-circularity
- [x] kore-non-terminating
- [x] kore-all-path-next-nt
- [x] kore-all-path-eventually
- [x] kore-all-path-rewrites
- [x] kore-all-path-rewrites-star
- [x] kore-all-path-rewrites-plus
- [x] kore-is-sort
- [x] ~~kore-inj~~  (_no longer needed_)
- [x] kore-is-predicate
- [x] kore-is-nonempty-sort

Note: It would probably be best to review the final diff, rather than individual commits.

Closes #749


generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,254 @@ def kore_forall(var: int) -> Notation:
...
+kore_weak_eventually = Notation(
+    'kore-weak-eventually',
+    2,
+    kore_not(
+        phi0_fv0,
+        kore_mu(svar0_id)(
+            phi0_fv0,
+            (kore_not(phi0_fv0, kore_or(phi1_fv0, (kore_next(kore_not(phi0_fv0, svar0)))))),
+        ),
+    ),
```
MirceaS, line 214: Can't this be defined in terms of `nu`?

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,254 @@ def kore_forall(var: int) -> Notation:
...
+kore_always = Notation(
+    'kore-always',
+    2,
+    kore_not(
+        phi0_fv0,
+        kore_mu(svar0_id)(
+            phi0_fv0,
+            (kore_not(phi0_fv0, kore_and(phi1_fv0, (kore_all_path_next(phi0_fv0, kore_not(phi0_fv0, svar0)))))),
+        ),
+    ),
```
MirceaS, line 230: same as above

MirceaS (dismissed):

generation/src/proof_generation/proofs/kore.py
malturki: I haven't thought carefully about that. It probably is expressible using `nu` instead. My idea was to stick to the definitions given in `kore.mm` as much as possible.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+    'kore-all-path--eventually',
+    2,
+    kore_mu(svar0_id)(phi0_fv0, kore_or(phi1_fv0, kore_all_path_next_nt(phi0_fv0, svar0))),
+    '(kG♢{1}):{0}',
+)
+
+# Notation ( \kore-all-path-rewrites ph0 ph1 ph2 ) ( \kore-implies ph0 ph1 ( \kore-all-path-next-nt ph0 ph2 ) )
+""" kore_all_path_rewrites(sort, left, right) """
+kore_all_path_rewrites = Notation(
+    'kore-all-path-rewrites', 3, kore_implies(phi0, phi1, kore_all_path_next_nt(phi0, phi2)), '({1} kG=> {2}):{0}'
```
fiedlr, line 345:
```suggestion
    'kore-all-path-rewrites', 3, kore_implies(phi0, phi1, kore_all_path_next_nt(phi0, phi2)), '({1} kA=> {2}):{0}'
```

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+)
+
+# Notation ( \kore-all-path-rewrites-star ph0 ph1 ph2 ) ( \kore-implies ph0 ph1 ( \kore-all-path-eventually ph0 ph2 ) )
+""" kore_all_path_rewrites_star(sort, left, right) """
+kore_all_path_rewrites_star = Notation(
+    'kore-all-path-rewrites-star',
+    3,
+    kore_implies(phi0_fv0, phi1, kore_all_path_eventually(phi0_fv0, phi2_fv0)),
+    '({1} kG=>* {2}):{0}',
```
fiedlr, line 354:
```suggestion
    '({1} kA=>* {2}):{0}',
```

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+)
+
+# Notation ( \kore-all-path-rewrites-plus ph0 ph1 ph2 ) ( \kore-implies ph0 ph1 ( \kore-all-path-next-nt ph0 ( \kore-all-path-eventually ph0 ph2 ) ) )
+""" kore_all_path_rewrites_plus(sort, left, right) """
+kore_all_path_rewrites_plus = Notation(
+    'kore-all-path-rewrites-plus',
+    3,
+    kore_implies(phi0_fv0, phi1, kore_all_path_next_nt(phi0_fv0, kore_all_path_eventually(phi0_fv0, phi2_fv0))),
+    '({1} kG=>+ {2}):{0}',
```
fiedlr, line 363:
```suggestion
    '({1} kA=>+ {2}):{0}',
```

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+    '({1} k=>+ {2}):{0}',
+)
+
+# Notation ( \kore-one-path-reaches-star ph0 ph1 ph2 ) ( \kore-implies ph0 ph1 ( \kore-weak-eventually ph0 ph2 ) )
+""" kore_one_path_reaches_star(sort, left, right) """
+kore_one_path_reaches_star = Notation(
+    'kore-one-path-reaches-star',
+    3,
+    kore_implies(phi0_fv0, phi1, kore_weak_eventually(phi0_fv0, phi2_fv0)),
+    '({1} k1=>* {2}):{0}',
```
fiedlr, line 294:
```suggestion
    '({1} kE=>* {2}):{0}',
```

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+)
+
+# Notation ( \kore-one-path-reaches-plus ph0 ph1 ph2 ) ( \kore-implies ph0 ph1 ( \kore-next ph0 ( \kore-weak-eventually ph0 ph2 ) ) )
+""" kore_one_path_reaches_plus(sort, left, right) """
+kore_one_path_reaches_plus = Notation(
+    'kore-one-path-reaches-plus',
+    3,
+    kore_implies(phi0_fv0, phi1, kore_next(kore_weak_eventually(phi0_fv0, phi2_fv0))),
+    '({1} k1=>+ {2}):{0}',
```
fiedlr, line 303:
```suggestion
    '({1} kE=>+ {2}):{0}',
```

generation/src/proof_generation/proofs/kore.py
fiedlr: `1` is confusing here. Once we run out of symbols, I would prefer `E` and `A` from CTL.

fiedlr (approved):
Looks good! Thanks, guys.

We can merge but please address my comments.
1. There's one freshness bug in one definition
2. I think we should be consistent with the ML paper when it comes to symbols. I edited the given symbols based on this + borrowed some notation from CTL whenever local and global symbols are used.
3. We need to update `kore_next` to also use • . The current diamond that we use is not used in the paper (I added it because it is common to use in modal logic).

generation/src/proof_generation/aml/syntax.py
```diff
@@ -659,8 +659,7 @@ class SSubst(Pattern):
     def __post_init__(self) -> None:
         # Check that SSubst is not redundant
         assert self.var != self.plug
-        # TODO: Add this check
-        # assert not self.pattern.s_fresh(self.var)
+        assert not self.pattern.svar_is_fresh(self.var.name), 'Redundant substitution: ' + self.pretty(PrettyOptions())
```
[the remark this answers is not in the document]
MirceaS: The `pattern` here shouldn't have freshness constraints. It's actually the opposite, we want it to not have freshness constraints because in almost all cases it should have `vars` in it in order for the SSubst to not be redundant, so I don't really get your point.
Sure, you can plug in a metavar that will end up not having the var when instantiated but then we can just throw away the SSubst at that point, this is just a quick post init check to weed out clearly redundant substs.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+        svar1,
+    ),
+)
+phi1_fv01 = MetaVar(
+    1,
+    s_fresh=(
+        svar0,
+        svar1,
+    ),
+)
```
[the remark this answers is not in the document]
MirceaS: Because it's only used once, so I don't wanna make a function more complex for a single isolated use case.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+
+
+def phi_pos(metavar: int, var: int) -> MetaVar:
```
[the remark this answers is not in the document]
MirceaS: Same reasoning as above

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+"""kore-well-founded(sort)"""
+kore_well_founded = Notation(
+    'kore-well-founded',
+    1,
+    kore_mu(svar0_id)(phi0_fv0, kore_all_path_next(phi0_fv0, svar0)),
+    'kWF:{0}',
+)
+
+
+# Notation ( \kore-well-founded-alt ph0 ) ( \kore-mu ph0 X ( \kore-all-path-next ph0 ( \kore-always ph0 X ) ) )
```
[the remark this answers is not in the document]
MirceaS: @malturki

generation/src/proof_generation/aml/syntax.py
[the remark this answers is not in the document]
fiedlr: This is unrelated to the PR. 

I agree with you, my point is that 

`assert not self.pattern.svar_is_fresh(self.var.name)`

is not the same as

`assert self.pattern.svar_is_free(self.var_name)`

that we cannot implement right now.

-----------------------------------------------------------

Why is it not the same?

`self.pattern.svar_is_free(self.var_name) => assert not self.pattern.svar_is_fresh(self.var.name)`

but not vice versa.

I can give you an example where 

`assert not self.pattern.svar_is_fresh(self.var.name) and not self.pattern.svar_is_free(self.var_name)`

Here it is:

`pattern := MetaVar(0, s_fresh=[])`

This satifies not-freshness criterion, but does not guarantee that `var_name` is free, simply instantiate with `sigma_1`. In this case, the substitution would be redundant.

So this check here is necessary, but insufficient to what we want to guarantee, that the subst is not redundant.

generation/src/proof_generation/proofs/kore.py
```diff
@@ -134,6 +140,271 @@ def kore_forall(var: int) -> Notation:
...
+
+
+evar_id = 0
+
+# Notation ( \kore-is-sort ph0 ) ( \exists-sort s0 ( \eq s0 ph0 ) )
+""" kore_is_sort(sort) """
+kore_is_sort = Notation(
+    'kore-is-sort',
+    1,
+    exists_sort(evar_id)(equals(EVar(evar_id), phi0)),
```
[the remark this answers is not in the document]
fiedlr: How about this?

generation/src/proof_generation/aml/syntax.py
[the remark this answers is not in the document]
MirceaS: Okay, yeah, I agree. Are you proposing we do anything about this?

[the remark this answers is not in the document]
fiedlr: We cannot do much about it now. We'll have to think about it though in the future.

PR #775, opened 2024-01-23
Add well-formedness for Circom v2 checker.

mcalancea:
Fixes [#777](https://github.com/runtimeverification/proof-checker/issues/777).

This PR adds well-formedness checking of string patterns for Circom v2 checker. 

During implementation I realized there's a simpler and more succinct circuit for doing the check (although its correctness is not more straightforward, it uses similar arguments). I prefer we merge this and if we feel the need later, I will implement the simple version and move this to `extras.circom`.


circom/src/v2/prop_h.circom
```diff
@@ -237,6 +246,87 @@ template GetLength(M) {
     res <== len;
 }
 
+// Returns 1 if pattern is syntactically valid, 0 otherwise
+// Requirements: 
+// - length input is obtained by GetLength()
+// - M > 1
```
traiansf, line 252: maybe say that `M` is the maximum pattern length

traiansf: maybe you could write a brief description of the well-formedness algorithm?

traiansf (approved):
Looks reasonable to me. I would have tried to also more incomplete patterns, but I guess it's not needed.

Would it be a good idea to write somewhere the sketch for the simpler well-formedness checker to make it easier to pick up if ever needed?

PR #776, opened 2024-01-23
Export proof to textual Hilbert style

traiansf:
The main purpose of this PR is to partially address Issue #779.

It creates objects containing the proof as a list of proof steps in a more traditional Hilbert-style.
Note that this is currently not optimized (at all); its main purpose is to make it easy to serialize it into Circom.

An elaborated output format of the proof is also provided.


generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,221 @@
...
+        claim_out: IO[Any] | None = None,
+        proof_out: IO[Any] | None = None,
+    ) -> None:
+        super().__init__(phase, out, claims, claim_out, proof_out)
+        self.steps: dict[str, int] = {}
+        self.index: int = 0
+        self.argument_index = 0
+        self.prop = 0
+        self.arguments: dict[str, HArgument] = {}
+        self.proof: list[HilbertStep] = []
```
nishantjr, line 132: @traiansf This storing of proofs will work for small proofs such as Propositional, but will likely get into trouble for longer executions. Would it be easy to "stream" the output so we don't need to do this?

nishantjr (commented):
Could you add a couple of tests? Perhaps just the output of `phi -> phi` intitially?

generation/src/proof_generation/interpreter/hilbert_interpreter.py
mcalancea: @nishantjr
We're aware of this issue in general, but I think for our current use cases it's a good compromise to store everything explicitly. It's less work and the Circom checker needs more insight to handle the proof rules that would make the larger proofs approachable anyway.

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,310 @@
...
+        self.steps.setdefault(str(ret.conclusion), self.index)
+        return ret
+
+    def publish_proof(self, proved: Proved) -> None:
+        super().publish_proof(proved)
+        self.hproof.set_arguments(list(self.arguments.values()))
+
+        self.out.write(f'{str(self.hproof)}\n')
+
+    def publish_axiom(self, axiom: Pattern) -> None:
```
mcalancea, line 300: This handling of axioms feels sketchy. If we don't handle any Hilbert proofs using axioms yet, let's raise an Error here and note to consider it more carefully once we do. 

The thing is that `publish_axiom` and `publish_claim` are executed in the Gamma/Claim phases and those aren't really proofs, but "constructions". Now you do indeed need some way to get the axioms you'll be using in the Proof phase inside the interpreter, but it's not clear to me that _this_ interpreter will always (or even should) run the Gamma phase beforehand. It makes sense to grab them (via constructor) from the `ProofExp` class instead I think?  

@nishantjr please comment on this.

mcalancea (changes requested):
Great work, Traian! Left some comments.

generation/src/proof_generation/interpreter/hilbert_interpreter.py
traiansf: I've added code to provide more intelligible messages for errors which would arise from the Axioms phase not being called here: https://github.com/Pi-Squared-Network/pi2/blob/74ed4295cced4ca05bb0fc5ca229d74aa4a9fd51/generation/src/proof_generation/interpreter/hilbert_interpreter.py#L193

MirceaS (approved):

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,285 @@
...
+        if self.prop == 1:
+            self.complete_delta(delta, 2)
+            self.hproof.add_proof_step(HProp1(self.index, ret.conclusion, delta[0], delta[1]))
+        elif self.prop == 2:
+            self.complete_delta(delta, 3)
+            self.hproof.add_proof_step(HProp2(self.index, ret.conclusion, delta[0], delta[1], delta[2]))
+        elif self.prop == 3:
+            self.complete_delta(delta, 2)
+            self.hproof.add_proof_step(HProp3(self.index, ret.conclusion, delta[0], delta[1]))
+        self.prop = 0
```
fiedlr, line 254:
```suggestion
        self.previous_step_prop_type = 0
```

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,285 @@
...
+        out: IO[Any],
+        claims: list[Claim] | None = None,
+        claim_out: IO[Any] | None = None,
+        proof_out: IO[Any] | None = None,
+    ) -> None:
+        super().__init__(phase, out, claims, claim_out, proof_out)
+        self.hproof = HilbertProof([], None, [])
+        self.steps: dict[str, int] = {}
+        self.index: int = 0
+        self.prop: int = 0
```
fiedlr, line 191:
```suggestion
        self.previous_step_prop_type: int = 0
```

fiedlr (changes requested):

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,285 @@
...
+        self.arg2 = arg2
+        self.idx_arg1 = idx_arg1
+        self.idx_impl = idx_impl
+
+    def pretty(self, opts: PrettyOptions) -> str:
+        return f'{self.index}: ModusPonens({self.arg1.pretty(opts)}, {self.arg2.pretty(opts)}, {self.idx_arg1}, {self.idx_impl}) -> {self.conclusion.pretty(opts)}'
+
+
+class HExistsQuantifier(HilbertStep):
+    """index: exists x, x"""
```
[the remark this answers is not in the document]
traiansf: Unfortunately I had it mixed with (Existence)...

mcalancea (commented):
Looks good to me, letting Adam approve.

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,292 @@
...
+    def pretty(self, opts: PrettyOptions) -> str:
+        return f'{super().pretty(opts)} by ExistsQuantifier({self.arg.pretty(opts)})'
+
+
+@dataclass(frozen=True)
+class HExistsGeneralization(HilbertStep):
+    """
+    idx_arg1: phi1 -> phi2
+    ------------------------------- EG
+    index: (exists arg2, phi1) -> phi2 (== conclusion)
```
fiedlr, line 97: `arg2 should be free in phi1`

This could be checked in `__post_init__`

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,292 @@
...
+    idx_arg1: int
+    arg2: EVar
+
+    def pretty(self, opts: PrettyOptions) -> str:
+        return f'{super().pretty(opts)} by ExistsGeneralization({self.idx_arg1}, {self.arg2.pretty(opts)})'
+
+
+class HAxiom(HilbertStep):
+    def __init__(self, index: int, conclusion: Pattern) -> None:
+        super().__init__(index, conclusion)
```
fiedlr, line 109: This does not need to be here, does it?

fiedlr: I guess it does not need to have the dataclass decorator, as it's not introducing new arguments

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,292 @@
...
+
+    def pretty(self, opts: PrettyOptions) -> str:
+        return f'{self.index}: Axiom({self.conclusion.pretty(opts)})'
+
+
+class HilbertProof:
```
fiedlr, line 115: I guess this could be an unfrozen dataclass

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,292 @@
...
+
+class ProofStepType(Enum):
+    NoStep = -1
+    Instantiate = 0
+    Prop1 = 1
+    Prop2 = 2
+    Prop3 = 2
+    ModusPonens = 4
+    ExistsQuantifier = 5
+    ExistsGeneralization = 6
```
fiedlr, line 170: Are these reflecting instruction numbers? Maybe they should eventually?

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,292 @@
...
+        right_idx = self.retrieve_proof_index(right.conclusion)
+        self.register_proof_step_conclusion(ret.conclusion)
+        self.hproof.add_proof_step(HModusPonens(self.index, ret.conclusion, left_idx, right_idx))
+        self.previous_step_type = ProofStepType.ModusPonens
+        return ret
+
+    def complete_delta(self, delta: dict[int, Pattern], size: int) -> None:
+        for i in range(size):
+            if i not in delta:
+                delta[i] = MetaVar(i)
```
fiedlr, line 241: Maybe this could actually be extracted to `utils.py` or something

generation/src/proof_generation/interpreter/hilbert_interpreter.py
```diff
@@ -0,0 +1,292 @@
...
+                        f'Could not retrieve index for pattern {str(p)}.',
+                        'Possible reasons:',
+                        '- missing an axiom (perhaps Gamma phase was not called);',
+                        '- An uninstantiated Prop step was used in the proof.',
+                    ]
+                )
+            )
+        return idx
+
+    def register_proof_step_conclusion(self, current_conclusion: Pattern) -> None:
```
fiedlr, line 210: I guess `register_proof_step_conclusion` should be called after adding to proof steps, and should take `current_conclusion` from the proof step where `index` is pointing (since `HilbertStep` has a conclusion too)

This will make it less likely to register something that does not match anything in the proof

fiedlr (approved):
Looks great! Thanks for addressing my suggestions. 

I just left a couple of minor ideas that might help and shouldn't be too hard to fix, but it's up to you if you want to implement them.

generation/src/proof_generation/interpreter/hilbert_interpreter.py
fiedlr: Not sure if this is doable in the current implementation though

traiansf: The way it's done now, the best way to check it would probably to have a check function in HilbertProof which goes over the entire proof and checks it for consistency.

mcalancea (approved):

PR #791, opened 2024-01-25
Convert semantics to the proof expression (part I)

IljaZakharov:
The main change is simply conversion of the semantics class to proof expression. Plus, it can now be directly imported along with notations.


generation/src/proof_generation/k/kore_convertion/language_semantics.py
```diff
@@ -555,6 +550,11 @@ def equational_rule(self, free_vars_pattern: Pattern, module: None | KModule = N
 
         self._equational_rule_base_substitutions[krule.ordinal] = base_substitutions
         return krule
+    
+    def collect_notations(self) -> None:
+        # the function should be called before the module is imported!
+        for symbol in self.symbols:
+            self.add_notation(symbol.app)
```
fiedlr, line 557: Why are we adding symbols as a notation?

fiedlr (approved):
Looks good!

generation/src/proof_generation/k/kore_convertion/language_semantics.py
IljaZakharov: Each symbol has a separate notation. Check out the `app` method of KSymbol class.

Issue #757, opened 2024-01-19
Kore Features To Eliminate/Simplify


IljaZakharov:
This is a high-level issue for ongoing discussions regarding a couple of Kore feature that the proof generation team would like to eliminate/simplify in Kore. Potentially we might want to discuss them all with the K team to discuss potential solutions.

Guiding Principles
===============

### Principle 1: Do not make changes purely for the sake of improving Kore.

It is important to remember that Kore has been deployed as is already in use by the various backends. It takes coordination and effort to change things. We should make changes only if it improves the lives of some consumer of Kore.  At the same time, we should make notes about the general direction we want to  move in.

### Principle 2:  Prefer  abstraction over precise matching logic definitions

The Kore language is intended to communicate between the frontend, backends, and various other tools the language semantics defined in a K definition in a succinct way. As such, it should not get bogged down by the details of matching logic. This unnecessarily forces users to reconstruct information. 
Matching logic supports notations. We should use these to allow abstraction. 

Concrete Issues
=============

Subsorting and Injections
--------------------------------------

The `definition.kore` file defines that `Int` is a subsort of `Tree` with the following axiom:

```
  axiom{R} \exists{R} (Val:SortTree{}, \equals{SortTree{}, R} (Val:SortTree{}, inj{SortInt{}, SortTree{}} (From:SortInt{}))) [subsort{SortInt{}, SortTree{}}()] // subsort
```

This axiom is quite vague, merely saying that `inj{Int, Tree}` is a function from `Int` to `Tree`.
I doubt that any backend  would find it useful to parse this axiom, and is likely just using the attribute.
In general, `inj` is quite underspecified. There are no axioms stating that if `Foo` is *not* a subsort of `Bar`, then `inj{Bar, Foo}` is bottom. In Matching Logic models,  `inj` may be an arbitrary function or partial function, or really many other arbitrary relations.

@nishantjr proposes we replace this with:

```
axiom{R} subsort{R, SortInt{}, SortTree{}}() [subsort{SortInt{}, SortTree{}}()]
```

Here,
*  `subsort{R, SortInt{}, SortTree{}}()` is documented to be matching logic notation for `subset{R} (Top{SortInt{}}, Top{SortTree{}})`.
[...]

PR #792, opened 2024-01-25
Fix base substitutions and make `tree-reverse-int` work

fiedlr:
Fixes a bug with base substitutions and adds the `tree-reverse-int` benchmark that is finally working now.

*Notes*
This removes incorrect substitutions from base substitutions. Base substitutions are only about mapping let-bound variables to their definitions. The substitutions removed in this PR are part of the hints, and they should not be part of base substitutions by definition. We were receiving incorrect patterns once more auxiliary variables (that are not let-bound) were introduced, which is the reason why proof generation failed for the `tree-reverse-int` example. See https://github.com/runtimeverification/proof-checker/issues/757#issuecomment-1910612107 . No need arised to treat injections as special in this example.


IljaZakharov (approved):
Looks super!

PR #799, opened 2024-01-26
Add version instruction

fiedlr:


docs/proof-language.md
```diff
@@ -833,6 +833,19 @@ Otherwise, execution aborts, and verification fails.
 `Pop`
 :   Consume the top of the stack.
 
+### Metadata
+
+`Version <n:u8> <m:u8>`
+:   Checks whether the proof checker supports the major version given by `n`, panics otherwise. [^version_declaration]
```
IljaZakharov, line 839: Have we agreed upon panic or warning? It is a warning in my notes.

IljaZakharov (approved):
Looks good. The only question is about panicking or issuing a warning if the version is not supported.

docs/proof-language.md
Robertorosmaninho: Are these `[^version_declaration]` on purpose? @fiedlr
I don't recall seeing something like this before. Sorry for asking late, but what do they mean?

PR #805, opened 2024-01-29
Welformedness checker, Noir style

traiansf:


mcalancea (approved):
Great to have this running so early, thanks Traian!

PR #807, opened 2024-01-29
Add more of the supported benchmarks

fiedlr:
Following the test based milestones (https://github.com/Pi-Squared-Network/pi2/issues/521), this PR adds more of the supported proofs to the test suite:

- `projection`
- `decrement`
- `builtin-int`
- `builtin-functions`: we add versions with an initial rewrite, which makes all work except for `abs` and `head-bytes`

Note: This PR adds `next-*` versions of inputs for `builtin-functions`, because proof generation for pretraces is not supported yet.


Robertorosmaninho (approved):
LGTM

malturki (approved):
Looks good. Just a heads up that the README and the test milestones issue may need to be updated.

PR #812, opened 2024-01-30
Semantics pe part ii

IljaZakharov:


generation/src/proof_generation/k/kore_convertion/language_semantics.py
```diff
@@ -549,8 +554,37 @@ def equational_rule(self, free_vars_pattern: Pattern, module: None | KModule = N
...
+                ongoing_proof.conc == rule.pattern
+            ), f"The proof doesn't prove the claim: {self.pretty_diff(ongoing_proof.conc, rule.pattern)}"
+
+            self.add_proof_expression(ongoing_proof)
+            return ongoing_proof
+        else:
+            # If it is already proven, we can just return the proof
+            expressions = filter(lambda exp: exp.conc == rule.pattern, self._proof_expressions)
+            required_one = next(expressions)
+            return required_one
```
fiedlr: Why? This is not needed, right? It won't prevent re-evaluation

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -427,8 +426,11 @@ def rewrite_event(self, rule: KRewritingRule, substitution: dict[int, Pattern])
             lhs == self.current_configuration
         ), f'The current configuration {self.pretty(self.current_configuration)} does not equal {self.pretty(lhs)}, the lhs of an instance of rule {self.pretty(rule.pattern)}'
 
+        # Get the proof for convenient rule representation ruled out from the initial rule axiom
+        rewrite_rule_proof = self.language_semantics.proof_for_rewrite_rule(rule)
+
         # Prove the instance of the used rule axiom
-        rewrite_step_proof = ExecutionProofExp.prove_substitutions(self, self.load_axiom(rule.pattern), substitution)
+        rewrite_step_proof = ExecutionProofExp.prove_substitutions(self, rewrite_rule_proof, substitution)
```
fiedlr: I take this is temporary, as we do want to load the axiom but in the simplified form?

generation/src/proof_generation/k/kore_convertion/language_semantics.py
```diff
@@ -486,33 +489,39 @@ def convert_ksort(
...
-                                assert hasattr(pattern, 'sort')
-                                converted_sort = semantics._convert_sort(scope, pattern.sort)
-                                converted_pattern = kl.kv(converted_sort, converted_pattern)
-                                eq_axiom = semantics.equational_rule(converted_pattern, module)
-                                semantics._cached_axiom_scopes[eq_axiom.ordinal] = scope
-                            else:
-                                next(module.counter)
+                            postponed_rule_imports.setdefault(kore_module.name, []).append(sentence)
+
+            # Stage two: Import rules
```
fiedlr, line 494: Why postpone them?

generation/src/proof_generation/proofs/kore.py
```diff
@@ -227,6 +229,51 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
...
+    """
+    phi
+    ---
+    phi[0 |-> substitution[0]]...[n |-> substitution[n]]
+    """
+    for name, plug in substitution.items():
+        assert_functional_pattern(semantics, plug)
+        prover.add_assumption(functional(0)(plug))
+        # TODO: Get rid of this hack:
+        subst_proofexp.add_axiom(func_subst_axiom(name))
```
fiedlr, line 261: Why is this not part of the substitution class then? and just reference with self?

generation/src/tests/unit/test_execution_proof_generation.py
```diff
@@ -827,42 +830,21 @@ def test_collect_functional_axioms() -> None:
 
 
 def test_simple_rules_pretty_printing() -> None:
```
fiedlr, line 832: test_tree_semantics_pretty_printing

generation/src/proof_generation/proofs/kore.py
```diff
@@ -231,7 +231,7 @@ def matching_requires_substitution(pattern: Pattern) -> dict[int, Pattern]:
 
 def assert_functional_pattern(semantics: LanguageSemantics, pattern: Pattern) -> None:
     """Check that the pattern is a functional symbol and it is valid to generate the axiom"""
-    if kore_dv.matches(pattern):
+    if isinstance(pattern, EVar) or kore_dv.matches(pattern):
```
fiedlr, line 234: So do we know how this is happening suddenly?

generation/src/proof_generation/k/execution_proof_generation.py
```diff
@@ -430,11 +357,14 @@ def rewrite_event(self, rule: KRewritingRule, substitution: dict[int, Pattern])
             lhs == self.current_configuration
         ), f'The current configuration {self.pretty(self.current_configuration)} does not equal {self.pretty(lhs)}, the lhs of an instance of rule {self.pretty(rule.pattern)}'
 
-        # Add the rule axiom
-        self.add_axiom(rule.pattern)
+        # Get the proof for convenient rule representation ruled out from the initial rule axiom
+        rewrite_rule_proof = self.language_semantics.proof_for_rewrite_rule(rule)
```
fiedlr, line 361: I would expect that these are added from claims, no need to run the proofs?

On the other hand, it might be a good sanity check.

fiedlr (approved):
Great work, Ilja. 

- First I thought I do not support running the proofs from `LanguageSemantics` in `ExecutionProofExp`, but it might be a good sanity check for debugging purposes. 

- I would just kindly ask for the remaining explanations, especially if you discovered how that `EVar` gets to `assert_functional_symbol`.

generation/src/proof_generation/k/execution_proof_generation.py
IljaZakharov: I like the idea about claims! I'll move proving to an earlier stage in the third PR anyway.

generation/src/proof_generation/k/kore_convertion/language_semantics.py
IljaZakharov: Some symbols aren't there yet, especially for instantiating the parametric sort for equational rules.

generation/src/proof_generation/proofs/kore.py
IljaZakharov: Yes, this is because we haven't checked base substitutions against these assertions.

generation/src/proof_generation/proofs/kore.py
IljaZakharov: Ok, let's move it there.

IljaZakharov: But it uses many things from the Kore theory, and we don't import anything from it to Substitution. I guess both places work. Let's discuss it in the third PR.

generation/src/tests/unit/test_execution_proof_generation.py
IljaZakharov: Applied

PR #823, opened 2024-01-30
Adding a RISC0 version of `sdc`

Robertorosmaninho:
This PR introduces the zk_sdc using the RISC0 backend.
It still doesn't have input support as we didn't define one.


sdc/src/main.rs
```diff
@@ -0,0 +1,16 @@
+#![deny(warnings)]
+use sdc::sdc;
+
+pub fn main() {
```
yzhang90, line 4: add a comment saying this file is just a place holder.

yzhang90 (approved):
LGTM. The only concern is the folder structure. The sdc and sdc_risc0 are both at the top level. @xc93

yzhang90:
please try it on larger input.

sdc/src/main.rs
Robertorosmaninho: Done!

PR #826, opened 2024-01-31
Add versioning system to the proof generator

fiedlr:
1) Introduces the `Version` instruction as given by `docs/proof-language.md`
    - This instruction is emitted at the beginning of a Gamma file 
    - This instruction is not mandatory for now, it just serves as a sanity check
2) Introduces a PR template to make sure PRs are in good shape and to remind bumping generator versions


Robertorosmaninho (approved):
LGTM! Thanks, Adam!

docs/pull_request_template.md
```diff
@@ -0,0 +1,17 @@
...
+- [ ] *All changes are appropriately tested*
+- [ ] `make check && make test-system` passes on my machine
+- [ ] I doublechecked `Commits` and `Files changed` and everything makes sense
+
+## Generation Subteam Checklist
+- [ ] the `pi2` module supports all the new changes
+- [ ] If I am changing the binary proof format itself, then
+  - [ ] the major/minor version is bumped in `generation/pyproject.toml` according to semver
+  - [ ] Rust checker in `rust/` is updated to support all the changes
+  - [ ] versions in `docs/proof-language.md`, `rust/Cargo.toml` and `rust/src/lib.rs` are in sync with `generation/pyproject.toml`
```
Robertorosmaninho, line 14: On second thought, I would add these two files as well:
   - [risc0/pi2/Cargo.toml](https://github.com/Pi-Squared-Network/pi2/blob/6241bdf97a84443e7bdc89c90e151ae37663105e/risc0/pi2/Cargo.toml#L3)
   - [risc0/pi2/guest/Cargo.toml](https://github.com/Pi-Squared-Network/pi2/blob/6241bdf97a84443e7bdc89c90e151ae37663105e/risc0/pi2/guest/Cargo.toml#L3)

Robertorosmaninho (changes requested):

docs/pull_request_template.md
fiedlr: Let's merge this please so that I do not have to update all the proofs again.

Robertorosmaninho (approved):

docs/pull_request_template.md
Robertorosmaninho: Sure, but any change on RISC0 should affect the versioning anyway, so it must always have the same version of the Rust Checker. Postpone this will only pass the problem to another person to solve.

Issue #824, opened 2024-01-30
Implement the zk versions of SDC


Robertorosmaninho:
- [x] Implement SDC in RISC0: #823 
- [ ] Implement SDC in Cairo: https://github.com/Pi-Squared-Network/pi2/pull/840

PR #840, opened 2024-02-02
feat: Introducing `sdc` in cairo

Robertorosmaninho:
Closes: #824

This PR presents the Cairo version of the SubProof Dependence Checker implemented in Rust.
The key differences are:
- Cairo doesn't support hashset, so we hash the elements to avoid collision and use an array
- Cairo doesn't support `rayon` or any other parallel library, so this version of `std` is sequential.

Other differences are only due to syntax/lib function differences! 

# PR Checklist

- [x] My PR is small (<= 15 commits)
- [x] Each commit is understandable, having a good description or self-explanatory changes
- [x] *All changes are appropriately tested*
- [x] `make check && make test-system` passes on my machine
- [x] I doublechecked `Commits` and `Files changed` and everything makes sense


yzhang90:
1. Cairo has hashset (i.e,[Felt252Dict<T>](https://github.com/starkware-libs/cairo/blob/main/corelib/src/dict.cairo)). Can we us it to mimic the hashset?
2. Does Felt252 impose the constraint on the length of the pattern?

Robertorosmaninho:
> 1. Cairo has hashset (i.e,[Felt252Dict](https://github.com/starkware-libs/cairo/blob/main/corelib/src/dict.cairo)). Can we us it to mimic the hashset?

Actually, this is a hashmap, I'm not sure what we can use as a "value" here since we expect the assumption to be the key and how much memory this can cost us. 
I would keep an eye on this option if we came across a performance issue due to array element access cost! As for the `Felt252Dict ` we need to create our own `insert`, `get`, and `contains` functions.

> 2. Does Felt252 impose the constraint on the length of the pattern?
Yes, each pattern string must have less than 31 char.

yzhang90 (approved):
LGTM

